PR #11, opened 2021-06-15
[Feature] Hemispherical distant sensor plugin

leroyvn:
## Description

This PR adds a hemispherical distant sensor, which is similar to an adjoint to the `envmap` emitter. It records radiance leaving the scene in a hemisphere. It can be useful to get a qualitative idea of the effective BRDF of a surface with complex geometry.

Code is mostly taken from the `distant` sensor plugin, and I don't know to what extent it would be desirable and/or doable to move common parts to a library header (in the end, only the constructor and direction sampling code are different).

An example of RGB output (a simple rectangular surface with a `roughconductor` BSDF; coloured dots each correspond to a directional emitter in the scene; note that pixel index origin is taken at the bottom left of the image):

Screenshot 2021-06-15 at 12 29 13

## To do

- [x] Improve docs with a example of output (also add up direction on it) and a schematic to explain film-hemisphere mapping
- [x] Update docs with actual test scene
- [x] Add tests

## Testing

A set of tests similar to those written for `distant` is included. We test for object construction, ray sampling and overall result correctness in a few scenarios. 

## Checklist

- [ ] My code follows the [style guidelines](https://mitsuba2.readthedocs.io/en/latest/src/developer_guide/intro.html#introduction) of this project
- [ ] I give permission that the Mitsuba 2 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba2/blob/master/LICENSE)


leroyvn:
After a quick discussion (https://github.com/mitsuba-renderer/mitsuba-data/pull/9#issuecomment-5193552084), we decided to close this PR, as `hdistant` seems quite domain-specific. For interested users, a more current implementation lives on the [Eradiate fork of Mitsuba](https://github.com/eradiate/mitsuba3/blob/main/src/eradiate_plugins/sensors/hdistant.cpp).

PR #40, opened 2021-10-11
Emissive Volume Rendering with Emitters

Microno95:
This PR adds emissive event sampling in the null-scattering `volpathmis` integrator for sampling the VRE and MIS for emission. It supersedes the previous PR #2 

## Description

Adds emissive volume support by adding a volume emitter. The interface, after discussion with @dvicini, involves adding an emitter attribute to medium which then behaves as a first class emitter (i.e. same as any other emitter). 

This change requires that endpoints accept an extra floating point random number in order to sample the volume. Aside from that, some machinery for transforming a SurfaceInteraction into a MediumInteraction and vice versa was incorporated to prevent bloat in the remainder of the code - currently sampling an emitter requires passing in a SurfaceInteraction which I did not want to change.

## Testing

- The updated codebase compiles with no new warnings.
- Automated python tests fail for LLVM modes for an unknown reason, I will investigate and update with a solution. Currently tests in `test_megakernel.py` fail with a segmentation error due to a missing reference to an LLVM variable - I'm open to suggestions for how to debug this.
- Most python tests pass, I'm working on fixing the issues with the remaining tests.
- Generated several emission test variants in `participating_media` which the integrator renders. These will probably need a separate PR for the mitsuba resources repo - additionally, these would need reference renders to be generated.
- Tested compilation for all `scalar_`, `llvm_`, `cuda_` single precision and double modes on both windows and ubuntu.

## Checklist:

- [x] My code also compiles for `cuda_*` and `llvm_*` variants.
- [x] I give permission that the Mitsuba 2 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba2/blob/master/LICENSE)

## Notes

- The implementation uses non-analogue probabilities as described in [Monte Carlo Methods for Volumetric Light Transport Simulation](https://cs.dartmouth.edu/wjarosz/publications/novak18monte.pdf) for 2 reasons:
	1. The radiance herein is assumed to not be the product of `sigma_a` and `L_e`, but rather simply `L_e` and while making the probability proportional to the radiance seems reasonable, I found that this led to fireflies in spectral modes and in fact setting the probability proportional to `ek::select(ek::hmean(radiance) > 0, sigma_t, 0.f)` leads to significantly better results.
	2. The scattering probability is left proportional to `sigma_t` for the reason that when proportional to `sigma_s` I was observing a reduced throughput in spectral rendering modes when the term `index_spectrum(sigma_s, channel)` was zero while non-zero for other channels.


johnson-c:
What is the status of this? Looks like it didn't make its way into mitsuba 3. Are there still plans to do this?

dvicini:
Not yet no. I am not sure if @Microno95 is currently actively working on this. We went back and forth on the implementation but there were still some correctness issues the last time we spoke.

Microno95:
I unfortunately had to put this on hold for the past two months due to paper writing and getting covid twice (yay). I'll be back to working on this soon, but as @dvicini said, there are correctness issues that need to be resolved first.

PR #116, opened 2022-07-07
Python wheels build with GitHub actions

njroussel:
## Description
This PR updates all the necessary code and scripts to build Python wheels using GitHub actions.

The GitHub action is setup to build `drjit` for the following configurations:
* Python versions 3.8, 3.9, 3.10
* Operating systems: Linux (x86_64), Windows (AMD64), macOS (x86_64, arm64)

:warning: The GitHub secrets might need to be updated (the tokens might have expired)

## Notes
These notes describe some of the limitations with the current setup.
- When cross-compiling for the macOS arm64 wheel, a native build must be compiled first. There are two reasons for this: we need to generate the spectral upsampling model and also generate the Python typing stubs. Both operations can only be done on a native build. The resulting files are then forwarded to the cross-compiled build. In addition, an arm64 version of the `drjit` package must be downloaded manually to resolve DrJit symbols during the Mitsuba build.
- Linux wheels are re-tagged manually to have the `manylinux` tag. We cannot use the `auditwheel` tool (which usually takes care of tagging the wheels as `manylinux`), because the `mitsuba` package will not be self-contained as it links to `drjit` shared libraries.
- Running tests inside of GitHub runners to validate a wheel is very unstable. This is why the GitHub action currently only tries to import the module and set a variant.
- `pyproject.toml` and `wheels.yml` both have hard-coded strings for the `drjit` version requirement. Automatically replacing these strings would require another script that would run earlier than the execution of those files.


jmuchovej:
Sorry if this is a necrobump, but is it possible to support `linux/arm64` variants, as well? I know that mainstream applications are unlikely to be run on arm64 Linuxes, but I've been packaging `mitsuba` in a devcontainer for a class. Having `linux/arm64` wheels would allow for the containers to run as `linux/arm64` on Macs, which sidesteps the need for Rosetta.

I'm happy to make the PR, I just don't know where to start since I'm actually a TA for this class (I don't use `mitsuba` in my day-to-day, so I'll need to read-up on how to do this).

jmuchovej:
Ooh. I see, this wouldn't be possible with GitHub Actions. https://github.com/actions/runner-images/issues/5631

Ideally, this is something that could be prioritized, though with the caveat that GitHub is currently blocking that realization of such wheels.

njroussel:
We could cross-compile. That's what we currently do to build the arm64 MacOS wheels.

However, I have no idea if there are any libraries that we depend on that can't be build for arm64. This might require some work. 

I must say, this is a very specific use case that I wouldn't even be able to fully-test myself (easily). I don't think we'd want to maintain something like this in the longer term. Unless there is more demand for it, of course.

If you do want to try this on your own, you should hopefully only need to modify the main `CMakeLists.txtl` and the GHA `wheels.yml`.

PR #138, opened 2022-07-26
Fixes for running dedicated built-in mi.Threads

tszirr:
Bugfix

## Description

Two of the three commits are simple Thread class fixes (no undefined default priorities, give back GIL while waiting for Thread methods to resume).

The third commit addresses a rather ugly conflict between `mi.Thread`-derived and `ThreadNotifier` `WorkerThread` instances. I am sure you are well aware of the somewhat unfortunate layering of threading functionality at this point. The provided fix addresses too possible implementations of C++ `thread_local` storage:
* The least surprising (and likely originally expected) behavior of `ThreadNotifier` initialization on thread start (not actually happening on my system), in which case the fix now replaces the `WorkerThread` instance by the instance of the dispatched `mi.Thread`-derived object with proper reference counting and destruction.
* The more surprising (but correct and observed on my system) initialization-before-use behavior of `ThreadNotifier`, in which case the fix prevents correct `mi.Thread`-derived objects to accidentally be replaced with anonymous `WorkerThread` instances that easily lead to crashes when lacking the right ThreadEnvironment settings.

## Testing

Minimal repro (Ubuntu 20.04 LTS, Python 3.8):
```python
import mitsuba as mi
import drjit as dr

mi.set_variant('scalar_rgb')

class RenderThread(mi.Thread):
    def __init__(self, scene):
        super().__init__(name='RenderThread')
        self.scene = scene

    def run(self):
        mi.set_variant('scalar_rgb')
        self.img = mi.render(self.scene)

scene_desc = mi.cornell_box()
scene = mi.load_dict(mi.cornell_box())
render_thread = RenderThread(scene)
render_thread.start()

render_thread.join()
plt.imshow(render_thread.img)
plt.show()
```

## Checklist:

*Please make sure to complete this checklist before requesting a review.*

- ~[ ] I have made corresponding changes to the documentation~
- ~[ ] I have added tests that prove my fix is effective or that my feature works~
- [x] I give permission that the Mitsuba 3 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba2/blob/master/LICENSE)

### DIFF
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -115,7 +115,7 @@ struct Thread::ThreadPrivate {
     bool external_thread = false;
     bool critical = false;
     int core_affinity = -1;
-    Thread::EPriority priority;
+    Thread::EPriority priority = Thread::ENormalPriority;
     ref<Logger> logger;
     ref<Thread> parent;
     ref<FileResolver> fresolver;
--- a/src/core/thread.cpp
+++ b/src/core/thread.cpp
@@ -91,17 +91,20 @@ std::atomic<uint32_t> WorkerThread::m_counter{0};
 struct ThreadNotifier {
     ThreadNotifier() {
         // Do not register the main thread
-        if (m_counter > 0)
-            Thread::register_external_thread("wrk");
+        if (m_counter > 0 && !self)
+            registered_external = Thread::register_external_thread("wrk");
+        else
+            registered_external = false;
         m_counter++;
     }
     ~ThreadNotifier() {
-        if (self)
+        if (registered_external)
             Thread::unregister_external_thread();
         m_counter--;
     }
     void ensure_initialized() {}
     static std::atomic<uint32_t> m_counter;
+    bool registered_external;
 };
 
 std::atomic<uint32_t> ThreadNotifier::m_counter{0};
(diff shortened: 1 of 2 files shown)


dvicini:
Hi Tobias, 
Thanks for the PR. I will hold off with merging this until the rest of the team is back from vacation. I not super familiar with the nitty gritty details of threading in Python/Pybind11 and can't fully assess if we should merge this as is or not.
Delio

src/core/thread.cpp
```diff
@@ -392,14 +395,27 @@ void Thread::start() {
...
+            if (self_)
+                self_->d->running = false;
+        }
+    } worker_identity;
     self = this;
+    if (worker_identity.self_) {
+        worker_identity.self_->dec_ref();
+        notifier.registered_external = false;
+    }
+    else {
```
Speierers, line 411:
```suggestion
    } else {
```

Speierers (commented):
The two first commits look good to me.

Regarding the third commit, I will let Wenzel take a lot a this patch, he is more familiar with this part of the codebase.

wjakob:
I am closing this very old pull request -- sorry that it was never merged, Tobias.
This whole threading system was always somewhat of a mess, and we just recently removed it altogether. You can now use Python or C++ threads without having to do anything special.

PR #140, opened 2022-07-26
tabphase: Clarify direction conventions and fix incorrect PDF value

### LINKED ISSUE #134: ❔ other question Geometric Phase Function interface
## Summary

I am currently about to implement new phase functions, and I have a question regarding the interpretation of the (geometric) phase function interface.

I think, you evaluate $\theta$ in what I call *computer-graphics* convention (both vectors pointing away from the interaction point), but compute the sampled vector $\omega_o$ in *physics* $\theta'$ convention, is this correct?

g1211
[Adopted from PBR book](https://pbr-book.org/3ed-2018/Volume_Scattering/Phase_Functions#)

## Description

This is a consistent interpretation, when I look at the HG-phase function implementation in [MTS2](https://github.com/mitsuba-renderer/mitsuba2/blob/master/src/phase/hg.cpp).

You evaluate in *computer graphics-convention* according to 
$$p(\cos \theta) = \frac{1}{4\pi} \frac{1-g^2}{(1+g^2+2g(\cos\theta)) ^{3/2} }$$

But sample in *physics* convention
$$\cos \theta = \frac{1}{2g} (1+g^2 - (\frac{1-g^2}{1-g+2g  \xi}))$$

So I came up with this interpretation of the local geometry,
g1011

where you eval $\theta$ but sample $\theta'$ and the blue vectors are the local frame in which we sample from a spherical coordinate system with its pole aligned with n. Is this correct?

In [MITS3s HG](https://github.com/mitsuba-renderer/mitsuba3/blob/master/src/phase/hg.cpp) you still evaluate in  *computer-graphics* convention and sample $\theta$ in the *physics* convention, but now you flip $\cos \theta$ in line [83](https://github.com/mitsuba-renderer/mitsuba3/blob/98a8ecb2390ebf35ef5f54f28cccaf9ab267ea48/src/phase/hg.cpp#L83).

Is this the correct way to sample in *physics* convention ? Or cant we just operate with the inversely sampled $\cos \theta$ here?

Similarly in [tabphase](https://github.com/mitsuba-renderer/mitsuba3/blob/master/src/phase/tabphase.cpp), I assume that you sample consistently with the [eval()](https://github.com/mitsuba-renderer/mitsuba3/blob/98a8ecb2390ebf35ef5f54f28cccaf9ab267ea48/src/phase/tabphase.cpp#L91) methods from a *ContinuousDistribution*, why do you call eval with $-\cos \theta$ here? 
 
Thank you very much for clarifying everything, I could create a (modest) PR for this if necessary ..

leroyvn:
## Description

This PR fixes #134. It clarifies local frame orientation in the `tabphase` plugin and fixes an incorrect PDF value.

## Testing

I added a test checking that the conventions used for direction sampling and PDF value computation are consistent.

## Checklist

- [x] I give permission that the Mitsuba 3 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba2/blob/master/LICENSE)

### DIFF
--- a/src/phase/tabphase.cpp
+++ b/src/phase/tabphase.cpp
@@ -26,7 +26,7 @@ function of the cosine of the scattering angle.
 
 .. admonition:: Notes
 
-   * The scattering angle is here defined as the dot product of the
+   * The scattering angle cosine is here defined as the dot product of the
      incoming and outgoing directions, where the incoming, resp. outgoing
      direction points *toward*, resp. *outward* the interaction point.
    * From this follows that :math:`\cos \theta = 1` corresponds to forward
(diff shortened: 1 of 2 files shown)


dvicini:
I just rebased this onto the latest master. Let's wait for the CI to re-run before merging

leroyvn:
Okay; before you proceed, please let me check the PDF normalisation as well (I'll add a test for this as well).

leroyvn:
@dvicini After a double check, I don't think more changes are required, so if you think it's fine, we can go.

dvicini:
I am about to bump the version number to 3.0.1 and release updated pip packages. Do you think this change should go into that? It's only fixes to a single plugin, so there is almost not risk in including it (if it passes the CI)

leroyvn:
Yes, I think it can, things are definitely better than before and those changes should not break the work of anybody (expect by making results more correct!).

dvicini:
Okay great. I will just wait for the CI to finish running on the PR and then merge it.

PR #188, opened 2022-08-19
Fix flag combination issue

krisrst:
*Please add the labels (e.g. bug, feature, ..) corresponding to this PR*

## Description

When running mitsuba3 like this:

`./mitsuba3/build/mitsuba -t 1 -m cuda_rgb mitsuba3/resources/data/scenes/cbox/cbox-rgb.xml  -o test`

I get this error:

```
Caught a critical exception: [xml.cpp:1117] Error while loading "scenes/biconvex/cbox_biconvex_test.xml" (near line 1, col 1): could not instantiate scene plugin of type "scene": 
[xml.cpp:1117]   [scene_optix.inl:177] Optix configuration initialization failed! The OptiX module compilation did not complete succesfully. The module's compilation state is: 0x2360
```

Specifying -t 1 and -m cuda_rgb is not a valid combination per se (setting a CPU thread count + GPU rendering makes no sense!) BUT - when this is accidentally input it makes for some head scratching...

The problem is caused by `pool_set_size` in thread.cpp:

```
void Thread::set_thread_count(size_t count) {
    global_thread_count = count;
    // Main thread counts as one thread
    pool_set_size(nullptr, (uint32_t) (count - 1));
}
```

..when this function is called with count = 1, and rendering with any cuda* backend, then optix initialisation fails. E.g. -t = 2 works:

`./mitsuba3/build/mitsuba -t 2 -m cuda_rgb mitsuba3/resources/data/scenes/cbox/cbox-rgb.xml  -o test`

I do not expect this to be a good fix for the issue. But it does show what is happening. So maybe some option checking would be more suited?

### DIFF
--- a/src/mitsuba/mitsuba.cpp
+++ b/src/mitsuba/mitsuba.cpp
@@ -242,7 +242,6 @@ int main(int argc, char *argv[]) {
                 thread_count = 1;
             }
         }
-        Thread::set_thread_count(thread_count);
 
         while (arg_define && *arg_define) {
             std::string value = arg_define->as_string();
@@ -256,6 +255,22 @@ int main(int argc, char *argv[]) {
         mode = (*arg_mode ? arg_mode->as_string() : MI_DEFAULT_VARIANT);
         bool cuda = string::starts_with(mode, "cuda_");
         bool llvm = string::starts_with(mode, "llvm_");
+
+
+        /*
+         * Ugly workaround for the issue:
+         * pool_set_size with count = 0 AND
+         * cuda rendering makes optix init
+         * fail.
+         * */
+        if( cuda ){
+            if( thread_count == 1 )
+                Thread::set_thread_count( 2 );
+            else
+                Thread::set_thread_count(thread_count);
+        }else{
+            Thread::set_thread_count(thread_count);
+        }
 
 #if defined(MI_ENABLE_CUDA)
         if (cuda)


merlinND:
Hello @krisrst,

Thanks for bringing this up and the attempted workaround.
A better solution would be to understand the root cause: why wouldn't OptiX be able to initialize when the thread count is set to 1?

krisrst:
Hi @merlinND ! Yeah I agree, but since I have limited experience with how optix works, I thought it best to discuss with you and just show you what the end of the rope looked like.

The optix initialisation failure is `OPTIX_MODULE_COMPILE_NOT_STARTED` - it looks a bit strange, since everything optix related is compiled fine.

It is certainly connected to this `pool` in some way. What is that used for?

I can do some more digging around, but maybe you could share any thoughts?

Anyhow, as this flag combination in itself is kind of "invalid" imo, I would argue that it was enough to just not allow setting number of CPU threads if you are doing GPU rendering.

krisrst:
Oh, on a side note.. I combined the -t 1 and -m scalar/gpu_rgb flag all the time on mitsuba2. So whatever it is, it was introduced in mitsuba3.

-Robin

merlinND:
I'd say that setting the number of CPU threads makes sense regardless of the mode, since there are always some CPU operations such as scene loading going on. So it's good to know about this bug!

I don't have time to look into it right now, but perhaps @Speierers or @njroussel will (it doesn't sound very urgent though).
My intuition is that there could be a kind of conflict with a special case we introduced in scene loading in JIT modes.

krisrst:
Ah :-) That's nice to hear, I thought this was just nitpicking.

Don't worry, I'll have a look at it for you guys. 

-Robin

Speierers:
Hi @krisrst ,

It will be fixed in the next release: [dcbfa75](https://github.com/mitsuba-renderer/mitsuba3/commit/dcbfa75a436420d80de431aa4225ae889d5640ff)

Thanks for reporting this.

merlinND:
Nice, thanks @Speierers!

PR #206, opened 2022-08-25
Parser should raise exception when kwarg is not used

Speierers:
This PR adds a check on keyword parameters passed to the parser to make sure those are used at least once. The previous behavior is too error prone.

### DIFF
--- a/src/core/xml.cpp
+++ b/src/core/xml.cpp
@@ -433,14 +433,17 @@ static std::pair<std::string, std::string> parse_xml(XMLSource &src, XMLParseCon
         if (!param.empty()) {
             std::sort(param.begin(), param.end(),
                 [](const auto &a, const auto &b) -> bool {
-                    return a.first.length() > b.first.length();
+                    return std::get<0>(a).length() > std::get<0>(b).length();
                 });
+
             for (auto attr: node.attributes()) {
                 std::string value = attr.value();
                 if (value.find('$') == std::string::npos)
                     continue;
-                for (const auto &kv : param)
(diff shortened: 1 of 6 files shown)


src/core/xml.cpp
```diff
@@ -433,14 +433,19 @@ static std::pair<std::string, std::string> parse_xml(XMLSource &src, XMLParseCon
...
             for (auto attr: node.attributes()) {
                 std::string value = attr.value();
                 if (value.find('$') == std::string::npos)
                     continue;
-                for (const auto &kv : param)
-                    string::replace_inplace(value, "$" + kv.first, kv.second);
+                for (auto &p : param) {
+                    if (string::replace_inplace(value, "$" + std::get<0>(p), std::get<1>(p))) {
+                        std::get<2>(p) = true;
+                        break;
```
njroussel, line 446: I would remove this `break` statement. With it, users wouldn't be able to have something like:
```
 <rgb name="radiance" value="$r_rad, $g_rad, $b_rad"/>
```
It's a weird use case, I admit.

njroussel (approved):

src/core/xml.cpp
merlinND: If we allow this, we need to make sure that variable names are well validated (e.g. no comma in them).

njroussel: In that case, even if we don't allow it we should have variable name checking, no?
```
<rgb name="radiance" value="$r_rad, 6.5, 2.3"/>
```
This example can still break.

PR #234, opened 2022-09-05
bug fix - Implemented a new bisection algorithm

### LINKED ISSUE #233: Builtin kd-tree sometimes fails due to bisection algorithm
## Summary

The bisection algorithm used in the internal kd-tree implementation fails in certain situations

## System configuration

System information:

  OS: Ubuntu 21.10
  CPU: AMD Ryzen 9 5900X 12-Core Processor
  Python: 3.9.7 (default, Sep 16 2021, 13:09:58) [GCC 7.5.0]
  CUDA: 11.3.109
  LLVM: 13.0.0

  Dr.Jit: 0.2.1
  Mitsuba: 3.0.1
     Is custom build? True
     Compiled with: Clang 11.0.1
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        llvm_ad_rgb

## Description

The bisection algorithm used in the internal kd-tree has a hard failure criterion that raises a runtime error if the bisection is not finished after 100 iterations. The rationale behind this, as far as I've understood is that a bisection that halves the set of points with each step should take no more than 64 steps for 64-bit numbers. However due to the non-uniform distribution of floating point numbers on the number line this assumption can fail if the bisection target is close to zero.

Based on this blog entry, I have implemented a fix for this issue, with comparable performance. https://www.shapeoperator.com/2014/02/22/bisecting-floats/

If you wish, I can submit a PR with this fix immediately.

## Steps to reproduce

It is hard for me to provide a MWE, because I only encounter this problem in a procedurally generated scene with several million objects. 

If there are further questions, I can provide more information about this issue.

schunkes:
## Description

This PR introduces a different algorithm for the bisection in the builtin kd-tree. The original implementation would fail to converge in the pre-defined 100 iterations.

Fixes #233 

## Testing

I did run the available tests and no new failures appeared.

## Checklist


schunkes:
I expanded the documentation of the code and implemented your suggestions. I don't know how you handle PRs, but once you are satisfied I can rebase on master and squash my commits and you can merge.

schunkes:
Done :)

wjakob:
Please use `dr::reinterpret_array<ScalarUInt32>(..)` instead of `reinterpret_cast`

schunkes:
Also done :+1:

wjakob:
Sorry, I forgot to mention: without the pointer manipulation. 

For example,
```
ScalarUInt left_int = dr::reinterpret_array<ScalarUInt>(left);
```

instead of

```
ScalarUInt left_int = *dr::reinterpret_array<ScalarUInt*>(&left);
```

include/mitsuba/core/math.h
```diff
@@ -235,6 +235,45 @@ MI_INLINE Index find_interval(dr::scalar_t<Index> size,
     return dr::binary_search<Index>(1, size - 1, pred) - 1;
 }
 
+/**
+ * \brief This function does the middle search for bisect
```
wjakob, line 239: Edit
```
This function computes a suitable middle point for use in the \ref bisect() function"
```

include/mitsuba/core/math.h
```diff
@@ -235,6 +235,45 @@ MI_INLINE Index find_interval(dr::scalar_t<Index> size,
...
+ * numbers is halved on each iteration of the bisection.
+ *
+ * Note that this strategy does not work if the numbers have different sign.
+ * In this case the function always returns 0.0 as the middle.
+ */
+template <typename Scalar>
+Scalar middle(Scalar left, Scalar right) {
+    using ScalarUInt = dr::uint_array_t<Scalar>;
+
+    // if either left or right are infinite, return the sum
```
wjakob, line 254:
```
// Propagate invalid (infinities, NaN) back to the caller
```

include/mitsuba/core/math.h
```diff
@@ -235,6 +235,45 @@ MI_INLINE Index find_interval(dr::scalar_t<Index> size,
...
+
+    // always return zero if left and right have different signs
+    if (dr::sign(left) != dr::sign(right) && left != Scalar(0.0) && right != Scalar(0.0))
+        return Scalar(0.0);
+
+
+    // we reinterpret as UInt using the absolute value, so we store the sign
+    // to reapply after interpreting the result back to Float
+    bool negate = left < Scalar(0.0) || right < Scalar(0.0);
+    left = dr::abs(left);
```
wjakob, line 268: I don't understand the need for this. Won't the middle point be the same if these numbers are negative? What happens if you remove the `dr::abs` calls and also remove the ` negate ? -mid : mid` part at the end?

wjakob (commented):
A few more comments.

include/mitsuba/core/math.h
schunkes: I am not sure about the necessity of the absolute value, but since we divide by two, the most significant bit in `mid_int` will always be zero and therefore the sign information is lost and must be reapplied from `negate`.

Speierers:
This looks good to me now. @wjakob can we merge this? 

@schunkes by any chance did you run the test suite with this changes? The CI doesn't compile Mitsuba without Embree so we wouldn't know if this breaks some of the tests.

wjakob:
It looks good to me.

wjakob:
@schunkes, can you comment on the test suite?

schunkes:
I have to say, some tests failed on my machine, but they did even with the unchanged master branch. I did not investigate that further, but I can say that no new tests failed after my changes to the kd-tree.

Speierers:
@schunkes indeed some tests where failing already when enabling the kdtree. This should be fixed on master now: [d9f4c47](https://github.com/mitsuba-renderer/mitsuba3/commit/d9f4c47939d4e251a26e20f0146c01dc5d0eef8f)

Could you rebase you branch and run the test suite again to make sure that everything works properly? Then we should be able to merge this PR.

schunkes:
I will have to do that in a few days, as I am leaving for a vacation today. I will get back to you, when I'm done.

schunkes:
Okay, I rebased on the current state of master and ran the tests. It looks good on my side :+1:

Issue #228, opened 2022-09-01
Build Error：DLL load failed while importing drjit_ext


NorvinZhang2022:
use “cmake -B./build -H. -G "Visual Studio 16 2019" -A x64”  created projects
then use "cmake --build ./build --config Release"  to build mitsuba3
And got error:
[[=============================
 dlpack.cpp
  Generating Code...
     Creating library H:/DRTookit/mitsuba3/build4/ext/drjit/src/python/Debug/drjit_ext.lib and object H:/DRTookit/mitsuba3/build4/ext/drjit/src/python/Debug/dr
  jit_ext.exp
  drjit-python.vcxproj -> H:\DRTookit\mitsuba3\build4\Debug\python\drjit\drjit_ext.cp39-win_amd64.pyd
  Traceback (most recent call last):
    File "H:\DRTookit\mitsuba3\ext\drjit\resources\generate_stub_files.py", line 317, in <module>
      import drjit as dr
    File "H:\DRTookit/mitsuba3/build4/Debug/python\drjit\__init__.py", line 24, in <module>
      import drjit.drjit_ext as drjit_ext  # noqa
  ImportError: DLL load failed while importing drjit_ext: 鎵句笉鍒版寚瀹氱殑妯″潡銆?
C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241,5): error MSB8066: Custom build for
'H:\DRTookit\mitsuba3\build4\CMakeFiles\01ef742381369a506b0f8633b071b686\drjit_stub_file_generation.rule;H:\DRTookit\mitsuba3\src\python\CMakeLists.txt' exited
 with code 1. [H:\DRTookit\mitsuba3\build4\src\python\drjit_stub_file_generation.vcxproj]
]]===================================

System information:
  OS: Windows 10 pro 21H1
  CPU: Intel(R) Core(TM) i9-10900K CPU
  GPU: Nvidia 3080
  Python version: 3.9.12

  CUDA version: 11.7
  NVidia driver: 516.94

  Mitsuba version: github master
  Compiled with: MSVC 19.29.30140.0

NorvinZhang2022:
I can find "drjit_ext.cp39-win_amd64.pyd" in folder "build\Release\python\drjit"
[...]

PR #245, opened 2022-09-08
Simplify DLL requirements at runtime from Python

njroussel:
## Description

This PR is a followup to this issue https://github.com/mitsuba-renderer/mitsuba3/issues/228, and other reports of failing builds on Windows.

`mitsuba` has a `os.add_dll_directory` statement in its `__init__.py` file. Unfortunately, this functionality is not well supported by Conda on Windows (https://github.com/conda/conda/issues/10897 and #228). Also, these statements are only useful when building the project locally, I can even imagine this producing some unwanted side-effects when using the pre-built wheels.
We expect users to run one of the `setpath` scripts before using `mitsuba` as an executable or the Python package. This script will properly set the `PATH` and `PYHTONPATH` environement variables to solve the necessary DLL paths and package resolution. Essentially, the `os.add_dll_directory` statements become useless if `setpath` has been run.
This PR removes that bit of code.

Unfortunately, the stub files are generated during the CMake build, hence we don't expect users to have run the `setpath` script yet. Up until now, they relied on the `__init__.py` to provide the missing DLL search directories.
This PR modifies the stub file generation such that the appropriate `PATH` and `PYTONPATH` environment is passed to it during the CMake build rather than relying on the package or the generation script itself to determine in it on it's own.


Speierers (commented):
Also looking good to me 👍

PR #246, opened 2022-09-09
rfilter.h: Include <limits>

tbaederr:
This file uses `std::numeric_limits`, so should include the standard <limits> header.

## Description

Fixes # (issue)

## Checklist

- [ ] I give permission that the Mitsuba 3 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba/blob/master/LICENSE)

### DIFF
--- a/include/mitsuba/core/rfilter.h
+++ b/include/mitsuba/core/rfilter.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <limits>
 #include <mitsuba/core/object.h>
 #include <mitsuba/core/math.h>
 #include <mitsuba/core/logger.h>


Speierers:
Is this necessary though? If your compiler complains I would be curious to know which compiler it is (and the version).

tbaederr:
This is necessary when using libstdc++, after applying https://github.com/mitsuba-renderer/nanogui/pull/130 as well.

Speierers:
When using Clang, Mitsuba should be compiled using `libc++` instead of `libstdc++`. Does the compilation succeed and the test suite run if you compile with `libstdc++` using your patch?

tbaederr:
> When using Clang, Mitsuba should be compiled using `libc++` instead of `libstdc++`.

I know. What i don't know is why that limitation exists.

> Does the compilation succeed and the test suite run if you compile with `libstdc++` using your patch?

The compilation succeeds. I am unfamiliar with pytest, so not sure how to properly run the test suite.  [The documentation](https://mitsuba2.readthedocs.io/en/latest/src/developer_guide/testing.html) just mentions invoking `pytest`, but that that just throws a `ModuleNotFoundError: No module named 'drjit'`.

Speierers:
you will need to run `source setpath.sh` first in order to set the Mitsuba and Dr.Jit PATH first. Then simply execute `python3 -m pytest ../src` from the `build` directory.

tbaederr:
Thanks for the instructions, that worked. I don't have a cuda-capable GPU so the tests using cuda all failed. However, both when using libstdc++ and libc++, 447 tests failed and 1413 tests succeeded (with `-m no slow`).

Speierers:
Okay I can merge this as it is a minor change.

PR #254, opened 2022-09-15
Differentiable shape `to_world` parameter

Speierers:
## Description

This PR attempts to enable the computation of gradients for the `to_world` parameter of the `sphere` analytical shape plugin.

The following changes are necessary to achieve this:
- Make the parameters update and state recomputation differentiable (e.g. when recomputing `m_radius`, `m_center`)
- Add support for the `DetachShape` and `FollowShape` flags in the `compute_surface_interaction` method.

This PR also implements `Sphere::eval_parameterization()` for texture sphere emitters.

### DIFF
--- a/src/shapes/sphere.cpp
+++ b/src/shapes/sphere.cpp
@@ -161,7 +161,7 @@ public:
 
     void traverse(TraversalCallback *callback) override {
         Base::traverse(callback);
-        callback->put_parameter("to_world", *m_to_world.ptr(), +ParamFlags::Differentiable);
+        callback->put_parameter("to_world", *m_to_world.ptr(), +ParamFlags::Differentiable | ParamFlags::Discontinuous);
     }
 
     void parameters_changed(const std::vector<std::string> &keys) override {
--- a/src/shapes/sphere.cpp
+++ b/src/shapes/sphere.cpp
@@ -338,10 +338,9 @@ public:
     ray_intersect_preliminary_impl(const Ray3fP &ray,
                                    dr::mask_t<FloatP> active) const {
         MI_MASK_ARGUMENT(active);
-        using Value = std::conditional_t<dr::is_cuda_v<FloatP> ||
-                                dr::is_diff_v<Float>,
-                            dr::float32_array_t<FloatP>,
-                            dr::float64_array_t<FloatP>>;
+        using Value = std::conditional_t<dr::is_cuda_v<FloatP> || dr::is_diff_v<Float>,
+                                         dr::float32_array_t<FloatP>,
+                                         dr::float64_array_t<FloatP>>;
         using Value3 = Vector<Value, 3>;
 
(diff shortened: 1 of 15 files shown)


src/shapes/sphere.cpp
```diff
@@ -402,35 +434,64 @@ class Sphere final : public Shape<Float, Spectrum> {
...
+                   get a point rigidly attached to the sphere attached point,
+                   including translation, scaling and rotation */
+                Normal3f n = dr::normalize(ray(pi.t) - center);
+                local = to_object.transform_affine(dr::fmadd(n, radius, center));
+                /* With FollowShape the local position should always be static as
+                   the intersection point follows any motion of the sphere. */
+                local = dr::detach(local);
+                si.p = to_world.transform_affine(local);
+                si.sh_frame.n = (si.p - center) / radius;
+                si.t = dr::sqrt(dr::squared_norm(si.p - ray.o) / dr::squared_norm(ray.d));
```
ziyi-zhang, line 479: Would this line leak the derivative of `ray` to `si.t` (with `FollowShape`)?

Speierers: This is fine as in the case where `ray` is moving, we care about the relative motion of the sphere and the ray (origin).

src/shapes/sphere.cpp
```diff
@@ -402,35 +434,70 @@ class Sphere final : public Shape<Float, Spectrum> {
...
+
+        const Point3f& center = m_center.value();
+        const Float& radius = m_radius.value();
+        const Transform4f& to_world = m_to_world.value();
+        const Transform4f& to_object = m_to_object.value();
+
+        /* If necessary, temporally suspend gradient tracking for all shape
+           parameters to construct a surface interaction completely detach from
+           the shape. */
+        dr::suspend_grad<Float> scope(detach_shape, center, radius, to_world, to_object);
```
ziyi-zhang, line 459: Throws error when trying to add a test for `DetachShape`

Speierers: This is fixed now in Dr.Jit

src/shapes/sphere.cpp
```diff
@@ -293,6 +307,24 @@ class Sphere final : public Shape<Float, Spectrum> {
...
+                                               uint32_t ray_flags,
+                                               Mask active) const override {
+        Point3f local = warp::square_to_uniform_sphere(uv);
+        Point3f p = dr::fmadd(local, m_radius.value(), m_center.value());
+
+        Ray3f ray(p + local, -local, 0, Wavelength(0));
+
+        PreliminaryIntersection3f pi = ray_intersect_preliminary(ray, active);
+
+        if (dr::none_or<false>(pi.is_valid()))
```
Speierers, line 320: Should take into account the rotation

Speierers:
Moved to #341

PR #283, opened 2022-09-27
OptiX 7.4 denoiser

njroussel:
## Description
This PR adds a wrapper for the Optix AI Denoiser.

The new `OptixDenoiser` class  has the following features:
* Denoise a noisy rendering without any extra information
* Denoising when also given the albedo and/or shading normals, to improve the denoising quality
* Temporal denoising, for which the optical flow and the previous frame (denoised) must also be provided

Important notes: 
* `box` filter must be used to produce the noisy rendering
* The shading normals must be in the coordinate frame of the sensor
* With temporal denoising, the optical flow can be set to zero and still produce convincing results

The `aov` integrator now also supports `albedo`. It actually only computes an approximation of the true albedo, as it will return the diffuse reflectance (`eval_diffuse_reflectance`) of the BSDFs.

## Testing
New tests are added for the `OptixDenoiser` to verify its implementation (`test_optixdenoiser.py`). In these tests, the denoised outputs are compared to references which are generated with the denoiser example in the OptiX SDK. (The `resources/data/tests/denoiser/denoiser_ref.py` script can generate these references)

### DIFF
--- a/include/mitsuba/render/bsdf.h
+++ b/include/mitsuba/render/bsdf.h
@@ -462,6 +462,9 @@ public:
     /// Set a string identifier
     void set_id(const std::string& id) override { m_id = id; };
 
+    /// Return the diffuse reflectance value (if any)
+    virtual Spectrum get_diffuse_reflectance(const SurfaceInteraction3f &si, Mask active = true) const;
+
     /// Return a human-readable representation of the BSDF
     std::string to_string() const override = 0;
 
@@ -541,6 +544,7 @@ DRJIT_VCALL_TEMPLATE_BEGIN(mitsuba::BSDF)
     DRJIT_VCALL_METHOD(eval_null_transmission)
     DRJIT_VCALL_METHOD(pdf)
     DRJIT_VCALL_METHOD(eval_pdf)
+    DRJIT_VCALL_METHOD(get_diffuse_reflectance)
     DRJIT_VCALL_GETTER(flags, uint32_t)
     auto needs_differentials() const {
         return has_flag(flags(), mitsuba::BSDFFlags::NeedsDifferentials);
--- a/include/mitsuba/render/bsdf.h
+++ b/include/mitsuba/render/bsdf.h
@@ -463,7 +463,8 @@ public:
     void set_id(const std::string& id) override { m_id = id; };
 
     /// Return the diffuse reflectance value (if any)
-    virtual Spectrum get_diffuse_reflectance(const SurfaceInteraction3f &si, Mask active = true) const;
+    virtual Spectrum get_diffuse_reflectance(const SurfaceInteraction3f &si,
+                                             Mask active = true) const;
 
     /// Return a human-readable representation of the BSDF
     std::string to_string() const override = 0;
--- a/include/mitsuba/render/bsdf.h
+++ b/include/mitsuba/render/bsdf.h
@@ -462,9 +462,22 @@ public:
     /// Set a string identifier
     void set_id(const std::string& id) override { m_id = id; };
 
-    /// Return the diffuse reflectance value (if any)
-    virtual Spectrum get_diffuse_reflectance(const SurfaceInteraction3f &si,
-                                             Mask active = true) const;
+    /**
+     * \brief Evaluate the diffuse reflectance
+     *
+     * This method approximates the total diffuse reflectance for a given
+     * direction. For some materials, an exact value can be computed
+     * inexpensively.
+     * When this is not possible, the value is approximated by
+     * evaluating the BSDF for a normal outgoing direction and returning this
+     * value multiplied by pi. This is the default behaviour of this method.
+     *
+     * \param si
+     *     A surface interaction data structure describing the underlying
+     *     surface position.
+     */
+    virtual Spectrum eval_diffuse_reflectance(const SurfaceInteraction3f &si,
+                                              Mask active = true) const;
 
     /// Return a human-readable representation of the BSDF
     std::string to_string() const override = 0;
@@ -545,7 +558,7 @@ DRJIT_VCALL_TEMPLATE_BEGIN(mitsuba::BSDF)
     DRJIT_VCALL_METHOD(eval_null_transmission)
     DRJIT_VCALL_METHOD(pdf)
     DRJIT_VCALL_METHOD(eval_pdf)
-    DRJIT_VCALL_METHOD(get_diffuse_reflectance)
+    DRJIT_VCALL_METHOD(eval_diffuse_reflectance)
     DRJIT_VCALL_GETTER(flags, uint32_t)
     auto needs_differentials() const {
         return has_flag(flags(), mitsuba::BSDFFlags::NeedsDifferentials);
--- a/include/mitsuba/render/optixdenoiser.h
+++ b/include/mitsuba/render/optixdenoiser.h
@@ -43,7 +43,7 @@ public:
      * \return A callable object which will apply the OptiX denoiser.
      */
     OptixDenoiser(const ScalarVector2u &input_size, bool albedo, bool normals,
-             bool temporal);
+                  bool temporal);
 
     OptixDenoiser(const OptixDenoiser &other) = delete;
 
@@ -60,20 +60,26 @@ public:
      * \param denoise_alpha
      *      Whether or not the alpha channel (if specified in the noisy input)
      *      should be denoised too.
+     *      This parameter is optional, by default it is true.
      *
      * \param albedo
      *      Albedo information of the noisy rendering.
-     *      (tensor shape: (width, height, 3))
+     *      This parameter is optional unless the OptixDenoiser was built with
+     *      albedo support. (tensor shape: (width, height, 3))
      *
      * \param normals
      *      Shading normal information of the noisy rendering. The normals must
      *      be in the coordinate frame of the sensor which was used to render
-     *      the noisy input. (tensor shape: (width, height, 3))
+     *      the noisy input.
+     *      This parameter is optional unless the OptixDenoiser was built with
+     *      normals support. (tensor shape: (width, height, 3))
      *
-     * \param normals_transform
+     * \param to_sensor
      *      A \ref Transform4f which is applied to the \c normals parameter
      *      before denoising. This should be used to tranform the normals into
      *      the correct coordinate frame.
+     *      This parameter is optional, by default no transformation is
+     *      applied.
      *
      * \param flow
      *      With temporal denoising, this parameter is the optical flow between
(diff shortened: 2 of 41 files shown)


include/mitsuba/render/bsdf.h
```diff
@@ -462,6 +462,10 @@ class MI_EXPORT_LIB BSDF : public Object {
     /// Set a string identifier
     void set_id(const std::string& id) override { m_id = id; };
 
+    /// Return the diffuse reflectance value (if any)
+    virtual Spectrum get_diffuse_reflectance(const SurfaceInteraction3f &si,
```
Speierers, line 466: Should we rename this method `diffuse_reflectance` or `eval_diffuse_reflectance`? Btw should this compute the diffuse reflectance or the albedo of the BSDF? IIRC the albedo can be computed as the integration of the diffuse reflectance for all incoming directions).

include/mitsuba/render/optixdenoiser.h
```diff
@@ -0,0 +1,188 @@
...
+     *      denoiser.
+     *
+     * \param normals
+     *      Whether or not shading normals information will also be given to the
+     *      Denoiser.
+     *
+     * \return A callable object which will apply the OptiX denoiser.
+     */
+    OptixDenoiser(const ScalarVector2u &input_size, bool albedo, bool normals,
+             bool temporal);
```
Speierers, line 41: indentation

include/mitsuba/render/optixdenoiser.h
```diff
@@ -0,0 +1,188 @@
...
+     * \param albedo
+     *      Albedo information of the noisy rendering.
+     *      (tensor shape: (width, height, 3))
+     *
+     * \param normals
+     *      Shading normal information of the noisy rendering. The normals must
+     *      be in the coordinate frame of the sensor which was used to render
+     *      the noisy input. (tensor shape: (width, height, 3))
+     *
+     * \param normals_transform
```
Speierers, line 72: Should this be renamed into `to_sensor_transform` or `to_sensor`?

include/mitsuba/render/optixdenoiser.h
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * \param previous_denoised
+     *      With temporal denoising, the previous denoised frame should be
+     *      passed here.
+     *      For the very first frame, the OptiX documentation recommends
+     *      passing the noisy input for this argument.
+     *      (tensor shape: (width, height, 3 | 4))
+     *
+     * \return The denoised input.
+     */
```
Speierers, line 93: In the docstring above, note all parameters that are optional.

src/bsdfs/roughplastic.cpp
```diff
@@ -501,6 +501,11 @@ class RoughPlastic final : public BSDF<Float, Spectrum> {
         return { depolarizer<Spectrum>(value) & active, pdf };
     }
 
+    /// Return the diffuse reflectance value (if any)
```
Speierers, line 504: Could maybe remove this comment?

src/integrators/aov.cpp
```diff
@@ -199,6 +207,19 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
...
+                    BSDFPtr bsdf = si.bsdf(ray);
+                    Spectrum spec = bsdf->get_diffuse_reflectance(si, active);
+                    Color3f color;
+                    if constexpr (is_spectral_v<Spectrum>)
+                        color = spectrum_to_srgb(spec, ray.wavelengths, active);
+                    else 
+                        color = spec;
+                    *aovs++ = dr::select(si.is_valid(), color.r(), 0.f);
+                    *aovs++ = dr::select(si.is_valid(), color.g(), 0.f);
+                    *aovs++ = dr::select(si.is_valid(), color.b(), 0.f);
```
Speierers, line 220: Do we need the `dr::select()` statements here? in case `si.is_valid() = false` the `bsdf` pointer should be `NULL` hence the virtual function call `get_diffuse_reflectance()` should already return zeros for those entries.

src/render/bsdf.cpp
```diff
@@ -21,6 +21,13 @@ MI_VARIANT Spectrum BSDF<Float, Spectrum>::eval_null_transmission(
     return 0.f;
 }
 
+MI_VARIANT Spectrum BSDF<Float, Spectrum>::get_diffuse_reflectance(
+    const SurfaceInteraction3f &si, Mask active) const {
+    Vector3f placeholder = Vector3f(0.0f, 0.0f, 1.0f);
```
Speierers, line 26: Rename `placeholder` -> `wo`

src/render/python/bsdf_v.cpp
```diff
@@ -96,6 +101,10 @@ template <typename Ptr, typename Cls> void bind_bsdf_generic(Cls &cls) {
              [](Ptr bsdf, const SurfaceInteraction3f &si, Mask active) {
                  return bsdf->eval_null_transmission(si, active);
              }, "si"_a, "active"_a = true, D(BSDF, eval_null_transmission))
+        .def("get_diffuse_reflectance",
+             [](Ptr bsdf, const SurfaceInteraction3f &si, Mask active) {
+                 return bsdf->get_diffuse_reflectance(si, active);
+             }, "si"_a, "active"_a = true)
```
Speierers, line 107: Add the docstring macro `D()`

src/python/main_v.cpp
```diff
@@ -84,6 +84,9 @@ MI_PY_DECLARE(quad);
 // render
 MI_PY_DECLARE(BSDFSample);
 MI_PY_DECLARE(BSDF);
+#if defined(MI_ENABLE_CUDA)
```
Speierers, line 87: It isn't necessary to guard those as the `#if def` are already in `optixdenoiser_v.cpp` no?

src/render/optixdenoiser.cpp
```diff
@@ -0,0 +1,391 @@
+#include <mitsuba/render/optixdenoiser.h>
+#include <mitsuba/render/optix_api.h>
+#include <drjit-core/optix.h>
+
+NAMESPACE_BEGIN(mitsuba)
+
+template <typename Float, typename Spectrum>
```
Speierers, line 7: Use `MI_VARIANT`?

src/render/optixdenoiser.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+MI_VARIANT
+OptixDenoiser<Float, Spectrum>::OptixDenoiser(const OptixDenoiser &other)
+    : m_input_size(other.m_input_size), m_options(other.m_options),
+      m_temporal(other.m_temporal) {
+    init(other.m_input_size, other.m_temporal);
+}
+
+MI_VARIANT
+OptixDenoiser<Float, Spectrum>::OptixDenoiser(OptixDenoiser &&other) noexcept {
```
Speierers, line 43: Do we really need to define the move/copy constructor/assignment operator?

include/mitsuba/render/optixdenoiser.h
```diff
@@ -0,0 +1,188 @@
...
+    TensorXf operator()(const TensorXf &noisy,
+                        bool denoise_alpha = true,
+                        const TensorXf *albedo = nullptr,
+                        const TensorXf *normals = nullptr,
+                        const Transform4f *normals_transform = nullptr,
+                        const TensorXf *flow = nullptr,
+                        const TensorXf *previous_denoised = nullptr) const;
```
Speierers, line 100: Why use pointers and no const references here? For default values?

src/render/optixdenoiser.cpp
```diff
@@ -0,0 +1,391 @@
...
+    const ref<Bitmap> &noisy_, bool denoise_alpha, const std::string &albedo_ch,
+    const std::string &normals_ch, const Transform4f &normals_transform,
+    const std::string &flow_ch, const std::string &previous_denoised_ch,
+    const std::string &noisy_ch) const {
+    if (noisy_->pixel_format() != Bitmap::PixelFormat::MultiChannel) {
+        size_t noisy_tensor_shape[3] = { noisy_->height(), noisy_->width(),
+                                         noisy_->channel_count() };
+        TensorXf noisy_tensor(noisy_->data(), 3, noisy_tensor_shape);
+        TensorXf denoised;
+        denoised = (*this)(noisy_tensor, denoise_alpha);
```
Speierers, line 201: One line

src/render/optixdenoiser.cpp
```diff
@@ -0,0 +1,391 @@
...
+                                        m_state_size, &guide_layer, &layers, 1,
+                                        0, 0, m_scratch, m_scratch_size));
+
+    size_t shape[3] = { noisy.shape(0), noisy.shape(1), noisy.shape(2) };
+    return TensorXf(std::move(output_data), 3, shape);
+}
+
+MI_VARIANT
+ref<Bitmap> OptixDenoiser<Float, Spectrum>::operator()(
+    const ref<Bitmap> &noisy_, bool denoise_alpha, const std::string &albedo_ch,
```
Speierers, line 192:
```suggestion
    const ref<Bitmap> &noisy, bool denoise_alpha, const std::string &albedo_ch,
```

src/render/optixdenoiser.cpp
```diff
@@ -0,0 +1,391 @@
...
+    };
+
+    std::unique_ptr<TensorXf> albedo_tensor = nullptr;
+    setup_tensor(albedo_bmp, albedo_tensor, 3);
+    std::unique_ptr<TensorXf> normals_tensor = nullptr;
+    setup_tensor(normals_bmp, normals_tensor, 3);
+    std::unique_ptr<TensorXf> flow_tensor = nullptr;
+    setup_tensor(flow_bmp, flow_tensor, 2);
+    std::unique_ptr<TensorXf> prev_denoised_tensor = nullptr;
+    setup_tensor(prev_denoised_bmp, prev_denoised_tensor, noisy_channel_count);
```
Speierers, line 284: Minor: could this lambda return a pointer instead?

src/render/optixdenoiser.cpp
```diff
@@ -0,0 +1,391 @@
...
+    };
+    check_resolution(noisy, m_input_size);
+    if (m_options.guideAlbedo)
+        check_resolution(*albedo, m_input_size);
+    if (m_options.guideNormal)
+        check_resolution(*normals, m_input_size);
+    if (m_temporal) {
+        check_resolution(*flow, m_input_size);
+        check_resolution(*previous_denoised, m_input_size);
+    }
```
Speierers, line 373: Skip the if statements below by making the lambda do nothing when the input pointer is nullptr.

Speierers (commented):
Looking good overall. A few minor suggestions and comments.

src/render/optixdenoiser.cpp
njroussel: I doubt anybody is actually going to use them.

However, the implicit default copy operators would be wrong. The other option is to `= delete` them all.

src/render/optixdenoiser.cpp
njroussel: :+1:

include/mitsuba/render/bsdf.h
njroussel: :+1: Ok with the rename.

So, I'm often confused about this. My understanding is that the albedo is defined as the fraction of radiance over irradiance, but that it is usually approximated by using only the diffuse component. Also, it should be defined as a double integral over the hemisphere (both incident and outgoing direction), such that it is truly view independent.
However, this would then have to be implemented as a MC integration for most BSDFs if we wanted it to be accurate (we only have a few existing BSDFs that have a closed form solution for this double integral). So the further approximation is to only consider the normal outgoing direction. Maybe we should just pay the performance cost for the MC integration with an arbitrary number of samples that can be set in the `aov` integrator?

src/render/optixdenoiser.cpp
Speierers: Yes, I think it would be enough to use `= delete` here. Anyway, the denoiser inherits from `Object` and will be use with `ref<>` most of the time, in which case we are always moving around a pointer to the denoiser and never really copying/moving its internal data.

src/bsdfs/mask.cpp
```diff
@@ -220,6 +220,12 @@ class MaskBSDF final : public BSDF<Float, Spectrum> {
         return dr::clamp(m_opacity->eval_1(si, active), 0.f, 1.f);
     }
 
+    Spectrum eval_diffuse_reflectance(const SurfaceInteraction3f &si,
+                                      Mask active) const override {
+        Float opacity = eval_opacity(si, active);
+        return opacity * (m_nested_bsdf->eval_diffuse_reflectance(si, active));
```
Speierers, line 226: Not sure this is correct. I think the albedo should be independent of the opacity here. E.g. when the opacity is very low, with the current code the albedo would be almost black which seems incorrect.
```suggestion
        return m_nested_bsdf->eval_diffuse_reflectance(si, active);
```

Speierers (commented):
One single comment, otherwise this looks great to me!

Issue #287, opened 2022-09-28
Empty/black pixels when rendering a simple shape with spp=1


trancelestial:
Hi, thank you for the amazing work you have done with the mitsuba renderer. I have a question regarding a simplified case where I observed an unexpected result: 

When rendering (aov + direct integrator) a simple rectangle shape (both 'type': 'rectangle', and a ply rectangle composed of only two triangles) with diffuse bsdf, using a single sample per pixel (spp=1), box filter, and a single point light I observe black/empty pixels both on the boundary and inside of the rendered shape (and its depth...). You can see the zoomed-in image below:

![image](https://user-images.githubusercontent.com/22893239/192789390-76cfe31c-4de9-4ce4-8063-6985ec958c9b.png)

Is this behavior expected? From my understanding for each pixel, a ray should be cast and intersected with the object. This intersection should be valid for the (black) pixels inside of the object. Since there is only a single sample per pixel, the box filter should basically output a single sample of the incoming radience for each pixel's ray, which for the rays that intersect with the shape's surface should be non-zero. I just started working with mitsuba, and might be getting the inner machinery of it wrong, so I would be grateful for any clarification/help you can provide to resolve this issue. Let me know if I can provide more information on my setting.

## System configuration

System information:

  OS: Ubuntu 20.04.5 LTS
  CPU: Intel(R) Core(TM) i9-10850K CPU @ 3.60GHz
  GPU: NVIDIA GeForce RTX 3090
  Python: 3.9.7 (default, Sep 16 2021, 13:09:58) [GCC 7.5.0]
  NVidia driver: 510.47.03
  CUDA: 11.5.119
  LLVM: 12.0.0

  Dr.Jit: 0.2.2
  Mitsuba: 3.0.2
     Is custom build? False
     Compiled with: GNU 10.2.1
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        llvm_ad_rgb

Speierers:
Hi @trancelestial ,

Could you share the scene file so we can give this a try on our side?

trancelestial:
Hi @Speierers, I couldnt find an easy way to export the scene object to XML format, but here is a python dictionary of the scene https://pastebin.com/y1cwfnXH, I hope this works.

Speierers:
Thanks, I am looking into that issue.

trancelestial:
Hi @Speierers,

[...]

PR #294, opened 2022-09-29
Fix numerical instabilities with box filter splatting

Speierers:
## Description

When using `box` reconstruction filter, it can happen that samples get splatted on neighboring pixels due to numerical instabilities. This can be seen for instance when rendering a depth image with one sample per pixel (see issue #287).

In this patch, the integrators provide a splatting position that doesn't contain the random offset in order to prevent this problem.

### DIFF
--- a/src/python/python/ad/integrators/common.py
+++ b/src/python/python/ad/integrators/common.py
@@ -407,7 +407,10 @@ class ADIntegrator(mi.CppADIntegrator):
                 # Return a reparameterized image position
                 pos_f = ds.uv + film.crop_offset()
 
-        return ray, weight, pos_f, reparam_det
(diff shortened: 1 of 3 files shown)


njroussel:
Why do we need the extra argument in `render_block` and `render_sample` to know if it's a box filter?
We have the `sensor` in both methods, can't we just do `bool box_filter  = sensor->film()->rfilter()->is_box_filter();` in both directly ?

Speierers:
You are right I didn't realise we had access to the sensor in there actually.

njroussel:
:ok_hand:  This looks good to me now!

wjakob:
Late to the party: looks good to me as well!

PR #378, opened 2022-11-10
Lower required Nvidia GPU architecture

### LINKED ISSUE #182: [🐛 bug report] Device ran out of memory, could not allocate 512 bytes
## Summary
The GPU immediately runs out of memory when using the cuda backend.

## System configuration
- Platform: Windows 10
- Compiler: MSVC 14.32
- Python version: 3.9.13
- Mitsuba 3 version: 3.0.1
- Compiled variants:
    * `scalar_rgb`
    * `cuda_rgb`
- GPU: GTX 980 Ti (6GB of memory)

## Description
I have been using the llvm backend with no problem, but when I try to use the cuda backend, it says that the device has run out of memory. This is the command line full log for just setting the variant to 'cuda_rgb':

    https://pastebin.com/1Xwc9tBq

And when I just render a scene from the console, this is the log:

    https://pastebin.com/wQCjR9nw

The same error happens even if I just write cuda_ for the variant, instead of telling me that the variant does not exist:

    Microsoft Windows [Versión 10.0.19044.1889]
    (c) Microsoft Corporation. Todos los derechos reservados.
    
    C:\Users\Diego>mitsuba -m cuda_
    
    Caught a critical exception: jit_malloc(): out of memory! Could not allocate 512 bytes of device memory.
    
    C:\Users\Diego>

The steps to reproduce are in the logs

njroussel:
This PR fixes #182.
It updates the `drjit` submodule where we now properly check if asynchronous memory allocations are supported. It also modifies the `.ptx` generation for custom shapes so as to be supported by older Nvidia GPU architectures.

(Very few changes, PR opened to run the CI)

### DIFF
--- a/resources/ptx/Makefile
+++ b/resources/ptx/Makefile
@@ -9,7 +9,7 @@ all: optix_rt.ptx
 optix_rt.ptx: $(mts_optix_main) $(mts_include_folder)/* $(mts_shape_folder)/*
 	nvcc $(mts_optix_main) \
 		 -I $(mts_include_folder) -I $(mts_shape_folder) -I $(OPTIX_PATH) \
-		 -O3 -gencode arch=compute_61,code=compute_61 --ptx
+		 -O3 -gencode arch=compute_50,code=compute_50 --ptx
 
 clean:
 	rm -f optix_rt.ptx
--- a/resources/ptx/optix_rt.ptx
+++ b/resources/ptx/optix_rt.ptx
@@ -7,7 +7,7 @@
 //
 
 .version 7.6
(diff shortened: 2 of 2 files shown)


wjakob:
LGTM!

Speierers (commented):
🚀

wjakob:
One quick question: why new EXR files? Is it actually related to the change in minimum SM version? (that would seem suspect)

njroussel:
No, they are not related. I've removed that change and also rebased.

(The `resources/data` submodule is lagging on `master` I'll do that separately.)

PR #384, opened 2022-11-14
Remove make opaque

Speierers:
Remove `dr::make_opaque()` in plugins and update drjit submodule

### DIFF
--- a/src/bsdfs/pplastic.cpp
+++ b/src/bsdfs/pplastic.cpp
@@ -210,8 +210,6 @@ public:
             s_mean = m_specular_reflectance->mean();
 
         m_specular_sampling_weight = s_mean / (d_mean + s_mean);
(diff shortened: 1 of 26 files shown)


wjakob:
Let's extend the commit message with a bit more context for others who may be following the repository.

> Currently, plugin constructors contain lots of  ``dr::make_opaque(...)`` and ``dr::opaque<T>(...)`` calls. Their purpose is to avoid per-plugin state from spilling into generated code so that the caching and function deduplication optimizations of Dr.Jit can kick in. Following updates in Dr.Jit (commit ref.), this is no longer necessary, and all of this code can be simplified by removing ``dr::make_opaque`/``dr::opaque``. The system from now on automatically treats plugin state as opaque when recording virtual function calls.

PR #389, opened 2022-11-16
Use BSDF::eval_pdf_sample() in path.cpp

Speierers:
This PR adds the `eval_pdf_sample` method to `BSDF` and uses it in the `path.cpp` integrator.

This reduces the number of virtual function calls performed by the integrator and reduces the rendering time by ~20% in `cuda_*` modes.

More work needs to be done to integrate such changes in the Python integrators (e.g. `prb`) as we need fine control over the differentiation of the BSDF routines (and detach sampling).

### DIFF
--- a/include/mitsuba/render/bsdf.h
+++ b/include/mitsuba/render/bsdf.h
@@ -379,7 +379,7 @@ public:
                       const Vector3f &wo,
                       Mask active = true) const = 0;
 
-/**
+    /**
      * \brief Jointly evaluate the BSDF f(wi, wo) and the probability per unit
      * solid angle of sampling the given direction. The result from the evaluated
      * BSDF is multiplied by the cosine foreshortening term.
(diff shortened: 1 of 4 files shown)


include/mitsuba/render/bsdf.h
```diff
@@ -417,6 +417,40 @@ class MI_EXPORT_LIB BSDF : public Object {
                                                 const Vector3f &wo,
                                                 Mask active = true) const;
 
+    /**
+     * \brief Jointly evaluate the BSDF f(wi, wo), the probability per unit
+     * solid angle of sampling the given direction \c wo and importance sample
+     * the BSDF model.
```
wjakob, line 423: Perhaps add a bit of context after the break. How about:

> This is simply a wrapper around two separate function calls to `eval_pdf()` and `sample()`. The function exists to perform a smaller number of virtual function calls, which has some performance benefits on highly vectorized JIT variants of the renderer. (A ~20% performance improvement for the basic path tracer on CUDA)

src/integrators/path.cpp
```diff
@@ -199,10 +203,20 @@ class PathIntegrator : public MonteCarloIntegrator<Float, Spectrum> {
...
                 }
 
-                // Evaluate BSDF * cos(theta)
-                Vector3f wo = si.to_local(ds.d);
-                auto [bsdf_val, bsdf_pdf] =
-                    bsdf->eval_pdf(bsdf_ctx, si, wo, active_em);
+                wo = si.to_local(ds.d);
+            }
+
+            // ------ Evaluate BSDF * cos(theta) & sampling direction -------
```
wjakob, line 209: sampling -> sample

wjakob:
Looks great, just a few comments. Not sure what is up with the CI. (Maybe related to the new GPU 😨)

njroussel (approved):

PR #394, opened 2022-11-17
Support large launches with up to 2^32 threads on OptiX backend

wjakob:
This commit should be merged following
https://github.com/mitsuba-renderer/drjit-core/pull/33. It adjusts the error messages that catch overly large wavefronts to the new capabilities of the OptiX backend.

### DIFF
--- a/src/render/integrator.cpp
+++ b/src/render/integrator.cpp
@@ -192,31 +192,38 @@ SamplingIntegrator<Float, Spectrum>::render(Scene *scene,
         if (develop)
             result = film->develop();
     } else {
+        size_t wavefront_size = (size_t) film_size.x() *
+                                (size_t) film_size.y() * (size_t) spp_per_pass,
+               wavefront_size_limit =
+                   dr::is_llvm_v<Float> ? 0xffffffffu : 0x40000000u;
+
+        if (wavefront_size > wavefront_size_limit) {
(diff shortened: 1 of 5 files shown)


Speierers (commented):
🚀 LGTM

src/render/integrator.cpp
```diff
@@ -619,23 +627,28 @@ AdjointIntegrator<Float, Spectrum>::render(Scene *scene,
...
+                               (size_t) spp_per_pass;
+
+            Log(Warn,
+                "The requested rendering task involves %zu Monte Carlo "
+                "samples, which exceeds the upper limit of 2^%zu = %zu for "
+                "this variant. Mitsuba will instead split the rendering task "
+                "%zu smaller passes to avoid exceeding the limits.",
+                samples_per_pass, dr::log2i(wavefront_size_limit + 1),
+                wavefront_size_limit, n_passes);
+        }
```
Speierers, line 647: It might be worth abstracting this logic into a separate function and adding bindings as we will want to use this in the Python integrators as well.

wjakob: Hi Sebastien -- the Python side is likely going to be more complicated (we do not support the multi-pass rendering, and it is not 100 percent clear how this will interact with all the various differentiable rendering methods). If you do not mind, I would like to move forward with this at least for the command line frontend.

PR #400, opened 2022-11-22
Set correct CUDA contexts before OptiX operations

njroussel:
Some OptiX operations require the current CUDA context to be the same as the one which was used to create the `OptixDeviceContext` which is passed as an argument to said operations. This was not the case in Mitsuba so far.

This PR fixes this by using an RAII-style scoped guard to set and unset the appropriate CUDA context for OptiX.


Speierers (commented):
LGTM 🚀

PR #410, opened 2022-11-29
Major Dr.Jit update

wjakob:
This PR updates the revision of Dr.Jit to incorporate a few major improvements:

- Variables are now consistently ordered, which improves the effectiveness of the function de-duplication optimization

- The 4 billion variable limit is gone

- Scene loading can now occur in parallel in JIT modes

For details, see

- https://github.com/mitsuba-renderer/drjit-core/pull/38
- https://github.com/mitsuba-renderer/drjit-core/pull/39
- https://github.com/mitsuba-renderer/drjit-core/pull/40
- https://github.com/mitsuba-renderer/drjit-core/pull/41

### DIFF
--- a/src/core/xml.cpp
+++ b/src/core/xml.cpp
@@ -225,7 +225,7 @@ struct XMLObject {
     size_t location = 0;
     ref<Object> object;
     std::mutex mutex;
-    uint32_t scope;
+    uint32_t scope = 0;
 };
 
 enum class ColorMode {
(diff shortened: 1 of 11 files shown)


src/core/xml.cpp
```diff
@@ -1095,6 +1102,10 @@ static Task *instantiate_node(XMLParseContext &ctx,
         }
 
         try {
+            #if defined(MI_ENABLE_CUDA) || defined(MI_ENABLE_LLVM)
+            if (ctx.is_jit() && ctx.parallel)
+                jit_new_scope((JitBackend) ctx.backend);
+            #endif
```
Speierers, line 1108: I am wondering whether we should move the `jit_new_scope` in the `PluginManager::create_object()` function. Even when instantiating objects sequentially, it would be safer if all plugins have their own JIT scope. 🤔

wjakob: Sure, why not. By the way, a call to `jit_new_scope` is now mandatory whenever there is a parallel code block (which potentially calls `jit_new_scope`) followed by further computation on the main thread that references the variables created in the parallel region. Otherwise the order in the generated program won't make sense.

Speierers (approved):

PR #413, opened 2022-11-29
Coalesce ImageBlock writes in CUDA/OptiX mode

wjakob:
This commit adjusts ``ImageBlock`` defaults so that it will try to generate more coherent scatter-reductions. We previously already did this on LLVM, where it improves performance quite a bit. It will similarly help on CUDA/OptiX once https://github.com/mitsuba-renderer/drjit-core/pull/42 is merged.


Speierers (approved):
🚀

njroussel (approved):
LGTM

Issue #417, opened 2022-12-01
Defining envmap as optimizable gives different result


dorverbin:
## Summary

Rendering a scene with an envmap and then making it optimizable and setting it to the ground truth value gives a different result.

## System configuration

System information:

  OS: CentOS Linux release 7.9.2009 (Core)
  CPU: Intel(R) Xeon(R) CPU E5-2698 v4 @ 2.20GHz
  GPU: Tesla V100-SXM2-16GB
  Python: 3.8.13 (default, Mar 28 2022, 11:38:47) [GCC 7.5.0]
  NVidia driver: 515.65.01
  CUDA: 9.0.176
  LLVM: 0.0.0

  Dr.Jit: 0.2.2
  Mitsuba: 3.0.2
     Is custom build? False
     Compiled with: GNU 10.2.1
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        llvm_ad_rgb

## Description

Rendering a simple sphere with the default `conductor` BSDF gives significantly different results before and after setting the optimizer to optimize the envmap. The errors, often in the order of ~0.1-0.2, seem to be mostly at the poles, clearly visible at the center of the image (see attached difference image, multiplied by ~5).
![mitsuba_envmap_opt](https://user-images.githubusercontent.com/15837806/204960608-db270293-3db7-4af8-aeb0-11c1a3cbc0dd.png)

## Steps to reproduce

Minimal example creating a random envmap, defining a scene, and rendering twice (with and without the optimizer):

```
from PIL import Image
import numpy as np
import matplotlib.pyplot as plt
import mitsuba as mi
mi.set_variant('cuda_ad_rgb')

# Create random (but azimuthally-constant) envmap.
envmap = np.uint8(255.0 * np.random.rand(50, 1, 3)).repeat(100, 1)
Image.fromarray(envmap).save('temp.png')

# Create simple spherical scene.
scene_dict = {
    'type': 'scene',
    'integrator': {
        'type': 'path',
        'max_depth': 2,
        'hide_emitters': True
    },
    'sensor': {
        'type': 'perspective',
        'fov': 39.6,
        'near_clip': 0.1,
        'far_clip': 1000.0,
        'to_world': mi.ScalarTransform4f.look_at(
            origin=[0, 4, 0],
            target=[0, 0, 0],
            up=[0, 0, 1]
        ),
        'sampler': {
            'type': 'independent',
            'sample_count': 2048
        },
        'film': {
[...]

PR #440, opened 2022-12-13
Envmap: Remove alpha channel, fix indexing

njroussel:
This PR applies two changes to the `envmap` plugin:

1) The environemnt-map data used to be stored as a RGBA `PixelFormat` inside a `TensorXf`. In this PR, the alpha channel is removed as it was efffectively never used and could lead to some confusion.

2) There was an error in the indexing which caused #417 (line 242 in `envmap.cpp`), this is fixed now.


Speierers (approved):
LGTM!

wjakob:
There is likely a problem with the spectral modes now. They use 4 entries to store polynomial coeffciientsl (which gives you color) and a scale (which gives you intensity). One option would be to insert a `std::conditional_t` in the right place to use a 4D type in spectral mode only.

PR #503, opened 2023-01-25
Fix parallel scene loading with Python plugins

### LINKED ISSUE #457: Loading a scene with python plugins fails when using parallel loading
## Summary

The default behavior for scene loading recently changed to always run in parallel, which breaks when the scene contains python plugins.

## System configuration

System information:

  OS: Ubuntu 22.04
  CPU: AMD Ryzen 3990X
  GPU: NVIDIA 3080Ti
  Python version: 3.9.12
  LLVM version: 11.0
  CUDA version: 11.7
  NVidia driver: 515.65.01

  Mitsuba: `master/10d3514a0295cad4ac6d440c7ff326561c6da6a2`

## Description

scenes are now loaded in parallel by default, which causes an error when loading a plugin defined in python. Indeed, when the object constructor is called from a new thread, the mitsuba variant is not set on the python side, leading to the following error:

```bash
RuntimeError: ​[xml.cpp:1140] Error while loading "<string>" (near byte offset 24): could not instantiate bsdf plugin of type "dummy": ImportError: Before importing any packages, you must specify the desired variant of Mitsuba using "mitsuba.set_variant(..)".
[xml.cpp:1140] The following variants are available: scalar_rgb, scalar_spectral, cuda_ad_rgb, llvm_ad_rgb.
[xml.cpp:1140] You can also use the MI_DEFAULT_VARIANT environment variable to define a default variant.
```

I don't think this has ever worked, but the commit that "broke" this on my end is [that](https://github.com/mitsuba-renderer/mitsuba3/commit/f4c0db0bb48b1f36de30659082355316de1da863) one, as it changes the default choice for scene loading to parallel.

## Steps to reproduce

Run the following script:

```python
import mitsuba as mi
mi.set_variant('scalar_rgb')

print(mi.variant()) # scalar_rgb
class DummyBSDF(mi.BSDF):
    def __init__(self, props):
        print(mi.variant()) # None
        mi.BSDF.__init__(self, props)

mi.register_bsdf('dummy', DummyBSDF)

# This works
mi.load_dict({'type': 'dummy'})

# This crashes
mi.load_string("<scene version='3.0.0'><bsdf type='dummy'/></scene>")

```

njroussel:
This PR fixes #457.

When loading a scene in parallel which makes use of custom Python plugins, none of the worker threads have the appropriate thread-local variant information (with the exception of the main thread).
This would produce the error seen in #457, as the workers that need to instantiate Python plugins will not be able to successfully call the constructor.

This PR sets the thread-local information, if needed, whenever a Python plugin is instantiated in C++ (more specifically: whenever a Python Plugin is instantiated through its RTTI-`Class`  object).

### DIFF
--- a/include/mitsuba/python/python.h
+++ b/include/mitsuba/python/python.h
@@ -63,7 +63,21 @@ PYBIND11_DECLARE_HOLDER_TYPE(T, mitsuba::ref<T>, true);
             (void) new Class(                                                  \
                 name, #Name, variant,                                          \
                 [=](const Properties &p) {                                     \
-                    py::object o = constructor(&p);                            \
+                    /* The thread-local python variant information might not
+                    have been set on this thread */                            \
+                    py::gil_scoped_acquire gil;                                \
+                    py::module mi = py::module::import("mitsuba");             \
+                    py::object py_variant = mi.attr("variant")();              \
+                    if (py_variant.is(py::none()) ||                           \
+                        py_variant.cast<std::string>() != variant)             \
+                        mi.attr("set_variant")(variant);                       \
+                                                                               \
+                    py::object o;                                              \
+                    {                                                          \
+                        py::gil_scoped_release release;                        \
+                        o = constructor(&p);                                   \
+                    }                                                          \
+                                                                               \
                     ref<Name> o2 = o.cast<Name*>();                            \
                     o.release();                                               \
                     return o2;                                                 \
--- a/src/core/tests/test_xml.py
+++ b/src/core/tests/test_xml.py
@@ -411,3 +411,17 @@ def test30_invalid_parameter_name(variant_scalar_rgb):
         </scene>
         """)
     e.match('Invalid character in parameter name')
+
+
+def test31_python_plugins_parallel(variants_vec_backends_once_rgb):
+    class DummyBSDF(mi.BSDF):
+        def __init__(self, props):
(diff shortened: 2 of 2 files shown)


include/mitsuba/python/python.h
```diff
@@ -63,6 +63,15 @@ PYBIND11_DECLARE_HOLDER_TYPE(T, mitsuba::ref<T>, true);
             (void) new Class(                                                  \
                 name, #Name, variant,                                          \
                 [=](const Properties &p) {                                     \
+                    /* The thread-local python variant information might not
+                    have been set on this thread */                            \
+                    py::gil_scoped_acquire gil;                                \
```
wjakob, line 68: Looks good. I would add a nested scope to ensure that we only hold the GIL for the `py::*` operations, but not the subsequent plugin construction.

njroussel: @wjakob I don't think this works. When I tested it, it seemed like the the local thread state was being deleted when the GIL was released.
 
 This line in Pybind seems to confirm it: https://github.com/pybind/pybind11/blob/f9f00495a3f0ef6037c215c42bd2d919590ff11f/include/pybind11/gil.h#L101
 
Should I try to workaround this with the C API directly?

wjakob: https://stackoverflow.com/questions/26061298/python-multi-thread-multi-interpreter-c-api

PR #552, opened 2023-02-22
Fix emitter sampling in volume integrators

### LINKED ISSUE #538: NEE issue on medium boundary with volpath and volpathmis
## Summary

Next-event estimation on a medium boundary surface in both **volpath** and **volpathmis** is not working as expected due to a possible issue with the parameters passed to `sample_emitter`, which leads to a random walk starting again when the ray has supposedly already exited the medium. This results in a darker image.

## System configuration

  OS: Windows-10
  CPU: AMD64 Family 25 Model 33 Stepping 2, AuthenticAMD
  GPU: NVIDIA GeForce RTX 3090 Ti
  Python: 3.10.8 | packaged by conda-forge | (main, Nov 24 2022, 14:07:00) [MSC v.1916 64 bit (AMD64)]
  NVidia driver: 526.47
  CUDA: 11.8.89
  LLVM: 0.0.0

  Dr.Jit: 0.3.2
  Mitsuba: 3.1.1
     Is custom build? False
     Compiled with: MSVC 19.33.31630.0
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        cuda_ad_spectral

## Description

Next-event estimation on a medium boundary surface in both **volpath** and **volpathmis** call a function called `sample_emitter`. This function is being passed the medium inside the boundary, even when called on the surface of the boundary, after exiting the medium. As a result, flags such as `active_medium` end up in a different state from those in the caller ( the `sample` function ).
This leads to a random walk starting again when the ray has supposedly already exited the medium.

This is particularly problematic if a Dirac emission profile is used.

One tentative fix would be to pass a nullptr when calling sample_emitters during surface NEE. This would be done here:
https://github.com/mitsuba-renderer/mitsuba3/blob/0e497163e13a27d7afe5beb591f6fd7bac7ce80d/src/integrators/volpath.cpp#L281

I have confirmed this to be an issue by rendering the same scene using mitsuba 0.6 and comparing images from mitsuba 3 with both the fix and the original source code.

The tentative fix image is similar to the mitsuba 0.6 image, whilst the original source code produces a different result.

I have included exrs and test scenes for in the attached zip file.
[mitsuba_ticket.zip](https://github.com/mitsuba-renderer/mitsuba3/files/10664350/mitsuba_ticket.zip)
## Steps to reproduce

1. Extract zipfile
2. mitsuba box_m3.xml

njroussel:
Volume integrators do not properly handle emitter sampling when the medium's boundary shape is smooth. It currently continues to account for the medium's extinction even though the ray has escaped it. This PR fixes this issue.

## Description

During emitter sampling for a surface interaction, we should check whether the sampled emitter direction would make the ray escape its current medium. The code currently continues to assume that no matter the sampled direction, it will still be in the medium.

Fixes #538

## Testing

There currently is no test with a medium enclosed in a smooth BSDF: only `null` or `dielectric` BSDFs, so emitter sampling for surface interactions is never tested.
(I am currently writing one, hence the draft PR. Also, I want to let the CI run to see if this breaks any existing tests.

### DIFF
--- a/src/integrators/volpath.cpp
+++ b/src/integrators/volpath.cpp
@@ -324,10 +324,11 @@ public:
 
 
     /// Samples an emitter in the scene and evaluates its attenuated contribution
+    template <typename Interaction>
     std::tuple<Spectrum, DirectionSample3f>
-    sample_emitter(const Interaction3f &ref_interaction, const Scene *scene,
-                   Sampler *sampler, MediumPtr medium, UInt32 channel,
-                   Mask active) const {
+    sample_emitter(const Interaction &ref_interaction, const Scene *scene,
+                   Sampler *sampler, MediumPtr medium,
+                   UInt32 channel, Mask active) const {
         Spectrum transmittance(1.0f);
 
         auto [ds, emitter_val] = scene->sample_emitter_direction(ref_interaction, sampler->next_2d(active), false, active);
@@ -339,6 +340,11 @@ public:
         }
 
         Ray3f ray = ref_interaction.spawn_ray(ds.d);
+
+        // Potentially escaping the medium if this is the current medium's boundary
+        if constexpr (std::is_convertible_v<Interaction, SurfaceInteraction3f>)
+            dr::masked(medium, ref_interaction.is_medium_transition()) =
+                ref_interaction.target_medium(ray.d);
 
         Float total_dist = 0.f;
         SurfaceInteraction3f si = dr::zeros<SurfaceInteraction3f>();
--- a/src/integrators/volpathmis.cpp
+++ b/src/integrators/volpathmis.cpp
@@ -373,8 +373,9 @@ public:
         return { result, valid_ray };
     }
 
+    template <typename Interaction>
     std::tuple<WeightMatrix, WeightMatrix, Spectrum, DirectionSample3f>
-    sample_emitter(const Interaction3f &ref_interaction, const Scene *scene,
+    sample_emitter(const Interaction &ref_interaction, const Scene *scene,
                    Sampler *sampler, MediumPtr medium,
                    const WeightMatrix &p_over_f, UInt32 channel,
                    Mask active) const {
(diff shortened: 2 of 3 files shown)


njroussel:
To further prove the issue, I've built a scene with sensor looking at a homogeneous medium enclosed in a `roughdielectric` and an area emitter located behind the medium. Below are the outputs when enabling/disabling emitter sampling for surface interactions in `volpath`:

Without emitter sampling:
image

With emitter sampling:
image

Here's the full scene description:
```xml
<scene version="3.2.1">
    <default name="spp" value="512" />
    <integrator type="volpath">
        <integer name="max_depth" value="40" />
        <integer name="rr_depth" value="40" />
    </integrator>
    <sensor type="perspective">
        <float name="fov" value="39.3077" />
        <transform name="to_world">
            <lookat target="-0.0, 0.0, -1.0" origin="0.0, 0.0, 0.0" up="-0.0, 1, 0.0" />
        </transform>
        <sampler type="independent">
            <integer name="sample_count" value="$spp" />
        </sampler>
        <film type="hdrfilm">
            <integer name="height" value="512" />
            <integer name="width" value="512" />
            <rfilter type="box" />
            <string name="file_format" value="openexr" />
        </film>
    </sensor>
    <medium type="homogeneous" id="smoke">
        <rgb name="albedo" value="0.98821396, 0.95783133, 0.9250646" />
        <rgb name="sigma_t" value="0.1103, 0.166, 0.1935" />
        <phase type="isotropic"/>
        <boolean name="sample_emitters" value="false" />
        <float name="scale" value="1000" />
    </medium>
    <shape type="obj">
        <string name="filename" value="cube.obj" />
        <transform name="to_world">
            <translate z="-4.0" y="0.0" />
        </transform>
        <ref name="interior" id="smoke" />
        <bsdf type="roughdielectric">
            <string name="distribution" value="beckmann" />
            <rgb value="0.25 0.25 0.25" name="specular_reflectance" />
            <rgb value="0.63 0.44 0.34" name="specular_transmittance" />
            <float name="alpha" value="0.30" />
            <float name="int_ior" value="1.345000" />
        </bsdf>
    </shape>
    <shape type="disk">
        <transform name="to_world">
            <lookat target="0.0,  0.0,  0.0" origin="0.0, 0.0, 10.0" up="0,0,0" />
        </transform>
        <emitter type="area">
            <rgb name="radiance" value="1000" />
        </emitter>
    </shape>
</scene>
```
(The `volpath` code was manually changed to disable emitter sampling on surfaces - there is no built-in property to do it).


Clearly the emitter sampling is accounting for some extra (undesired) absorption. This is what is also expected by looking at the code.

Issue #544, opened 2023-02-16
Struct converter fails in non-vectorized mode


dvicini:
Hi, 

This is somewhat of a niche issue I ran into today: if Mitsuba is run without vectorization flags, the struct converter fails to properly convert the rendered image and the output image when rendering the Cornell box is black. This can be reproduced by running

```
cmake .. -GNinja -DMI_NATIVE_FLAGS=""
ninja
mitsuba  ../resources/data/scenes/cbox/cbox.xml

```

So there is some kind of inconsistency between AVX and non-AVX mode in struct.cpp. A few tests in test_struct.py fail as well, which could help localize the issue. 

I don't really have a need to fix this on our end for now, but it would probably be good to either fix this code path or just disable the struct JIT for non-avx machines (i.e., CPUs that are >10 years old at this point).

Best,
Delio

njroussel:
Hi @dvicini 

I've pushed a fix https://github.com/mitsuba-renderer/mitsuba3/commit/8e441f0361d6def2c6f101fd8243d8646d57b9be.

The issue was even "worse": we were using `asmjit` to generate `SSE` instructions which are also obviously not available. I'm actually a bit surprised that this runs just fine :exploding_head:.

dvicini:
Hi @njroussel 

Thanks for taking a look. Indeed, at least SSE instructions need to be available. However, I don't think that fixes the issue. I think the reason it worked anyway is that all that matters at runtime is the actual hardware, not the flags specified to the compiler. 

By running the following cmake command, the JIT compiler remains enabled (since SSE instructions are available), but doesn't use the AVX code path:
`cmake .. -GNinja -DMI_NATIVE_FLAGS="-march=westmere" `

So clearly something is broken in the SSE code path. I spend some time today trying to isolate it. The tests that fail are test14_weight, test16_alpha_1 and test19_alpha_4. I think it might be somehow related to some of the logic around L1344. If I comment out this part, the test14_weight at least passes. I couldn't quite isolate why this code would fail though.

njroussel:
I see, I wrongly assumed that `asmjit` would disable instructions depending on the flags used to compile.

I took a quick stab as this. Indeed, it seems related to the `blendvps` instruction (L1347 and L1391). Not immediately obvious what's wrong here. I'll take another look tomorrow.

njroussel:
[...]

PR #564, opened 2023-02-28
Upgrade asmjit

njroussel:
## Description

This PR is a tentative fix to #544. As shown in the issue, the current version of `asmjit` generates wrong assembly. This PR updates the submodule and makes the necessary (breaking changes) refactorings in `struct.cpp` to handle the new version. The new version passes all tests, even when compiled without AVX (only SSE4.2).


njroussel:
Note: The `load()` operation for 32 bit integers was wrong and was fixed in this PR too (there was no sign extension when using signed integers)

PR #589, opened 2023-03-14
[✨ feature] Implement `eval_roughness()` for most BSDFs

saeedhd96:
## Description

Introduce method `eval_roughness()` similar to eval_diffuse_reflectance() ...

Fixes #454

## Testing

Tested on a two-sided BSDF, the test is named `test04_eval_roughness` in `test_twosided.py`

## Checklist

### DIFF
--- a/include/mitsuba/python/docstr.h
+++ b/include/mitsuba/python/docstr.h
@@ -758,6 +758,13 @@ Parameter ``si``:
     A surface interaction data structure describing the underlying
     surface position.)doc";
 
+static const char *__doc_mitsuba_BSDF_eval_roughness =
+R"doc(Evaluate the roughness
+
+Parameter ``si``:
+    A surface interaction data structure describing the underlying
+    surface position.)doc";
+
 static const char *__doc_mitsuba_BSDF_eval_null_transmission =
 R"doc(Evaluate un-scattered transmission component of the BSDF
 
--- a/include/mitsuba/render/bsdf.h
+++ b/include/mitsuba/render/bsdf.h
@@ -519,6 +519,23 @@ public:
     virtual Spectrum eval_diffuse_reflectance(const SurfaceInteraction3f &si,
                                               Mask active = true) const;
 
+    /**
+     * \brief Evaluate roughness
+     *
+     * This method approximates the roughness for a given
+     * direction. For some materials, an exact value can be computed
+     * inexpensively.
+     * When this is not possible, the value is approximated by
+     * evaluating the BSDF for a normal outgoing direction and returning this
+     * value multiplied by pi. This is the default behaviour of this method.
+     *
+     * \param si
+     *     A surface interaction data structure describing the underlying
+     *     surface position.
+     */
+    virtual Float eval_roughness(const SurfaceInteraction3f &si,
+                                              Mask active = true) const;
+
     /// Return a human-readable representation of the BSDF
     std::string to_string() const override = 0;
 
@@ -600,6 +617,7 @@ DRJIT_VCALL_TEMPLATE_BEGIN(mitsuba::BSDF)
     DRJIT_VCALL_METHOD(eval_pdf)
     DRJIT_VCALL_METHOD(eval_pdf_sample)
     DRJIT_VCALL_METHOD(eval_diffuse_reflectance)
+    DRJIT_VCALL_METHOD(eval_roughness)
     DRJIT_VCALL_GETTER(flags, uint32_t)
     auto needs_differentials() const {
         return has_flag(flags(), mitsuba::BSDFFlags::NeedsDifferentials);
--- a/src/bsdfs/twosided.cpp
+++ b/src/bsdfs/twosided.cpp
@@ -281,6 +281,30 @@ public:
             return result;
         }
     }
+    Float eval_roughness(const SurfaceInteraction3f &si_,
+                                      Mask active) const override {
+        SurfaceInteraction3f si(si_);
+
+        if (m_brdf[0] == m_brdf[1]) {
+            si.wi.z() = dr::abs(si.wi.z());
+            return m_brdf[0]->eval_roughness(si, active);
+        } else {
+            Float result = 0.f;
+            Mask front_side = Frame3f::cos_theta(si.wi) > 0.f && active,
+                 back_side  = Frame3f::cos_theta(si.wi) < 0.f && active;
+
+            if (dr::any_or<true>(front_side))
+                result = m_brdf[0]->eval_roughness(si, front_side);
+
+            if (dr::any_or<true>(back_side)) {
+                si.wi.z() *= -1.f;
+                dr::masked(result, back_side) =
+                    m_brdf[1]->eval_roughness(si, back_side);
+            }
+
+            return result;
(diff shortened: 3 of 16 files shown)


include/mitsuba/python/docstr.h
```diff
@@ -754,6 +754,13 @@ inexpensively. When this is not possible, the value is approximated by
 evaluating the BSDF for a normal outgoing direction and returning this
 value multiplied by pi. This is the default behaviour of this method.
 
+Parameter ``si``:
+    A surface interaction data structure describing the underlying
+    surface position.)doc";
+
+static const char *__doc_mitsuba_BSDF_eval_roughness =
+R"doc(Evaluate the roughness
+
```
njroussel, line 763: Can you rerun the `docstrings` target please?  This is currently missing some of the documentation you wrote :/

src/render/python/bsdf_v.cpp
```diff
@@ -64,6 +64,10 @@ MI_VARIANT class PyBSDF : public BSDF<Float, Spectrum> {
                                       Mask active) const override {
         PYBIND11_OVERRIDE_PURE(Spectrum, BSDF, eval_diffuse_reflectance, si, active);
     }
+    Float eval_roughness(const SurfaceInteraction3f &si,
```
njroussel, line 67: Missing new line here.

include/mitsuba/render/bsdf.h
```diff
@@ -519,6 +519,23 @@ class MI_EXPORT_LIB BSDF : public Object {
...
 
+    /**
+     * \brief Evaluate roughness
+     *
+     * This method approximates the roughness for a given
+     * direction. For some materials, an exact value can be computed
+     * inexpensively.
+     * When this is not possible, the value is approximated by
+     * evaluating the BSDF for a normal outgoing direction and returning this
+     * value multiplied by pi. This is the default behaviour of this method.
```
njroussel, line 530: Please update this :smile:

src/render/bsdf.cpp
```diff
@@ -40,6 +40,11 @@ MI_VARIANT Spectrum BSDF<Float, Spectrum>::eval_diffuse_reflectance(
     return eval(ctx, si, wo, active) * dr::Pi<Float>;
 }
 
+MI_VARIANT Float BSDF<Float, Spectrum>::eval_roughness(
+    const SurfaceInteraction3f & /* si */, Mask /* active */) const {
+    return 1.f;
```
njroussel, line 45: Let's have the default value be 0.

njroussel (changes requested):
Thank you for this PR @saeedhd96 , especially for going through all of these BSDFs.

A part from the few minor comments, could I ask you to please add a test for the `aov` changes? Preferably in a new file `src/integrators/tests/test_aov.py` -- you only need to test the roughness. You can just spawn a single ray that is supposed to hit a rectangle with known roughness and verify the `aovs` return value. (We don't have any existing tests for the `aov` integrator.)

include/mitsuba/render/bsdf.h
```diff
@@ -519,6 +519,23 @@ class MI_EXPORT_LIB BSDF : public Object {
...
+     *
+     * \param si
+     *     A surface interaction data structure describing the underlying
+     *     surface position.
+     */
+    virtual Float eval_roughness(const SurfaceInteraction3f &si,
+                                              Mask active = true) const;
```
Speierers, line 537: Indentation?

src/bsdfs/twosided.cpp
```diff
@@ -281,6 +281,30 @@ class TwoSidedBRDF final : public BSDF<Float, Spectrum> {
...
+            if (dr::any_or<true>(front_side))
+                result = m_brdf[0]->eval_roughness(si, front_side);
+
+            if (dr::any_or<true>(back_side)) {
+                si.wi.z() *= -1.f;
+                dr::masked(result, back_side) =
+                    m_brdf[1]->eval_roughness(si, back_side);
+            }
+
+            return result;
```
Speierers, line 305: Could we move this logic to a separate function?

Speierers:
Thanks @saeedhd96 for this PR!

It would be beneficial if this feature could also take into account anisotropic roughness. Would it be best to have this method return two roughness values? Is there something I'm not considering?

include/mitsuba/python/docstr.h
saeedhd96: Sorry I am not sure I understand what you mean by rerunning `docstrings`?

include/mitsuba/render/bsdf.h
saeedhd96: Why shouldn't it be exactly the same? doesn't it under the hood call same methods as diffuse_reflectance_eval?

src/render/bsdf.cpp
saeedhd96: Wouldn't it cause wierd behavior for the downstream applications of this method? Isn't the default roughness for a diffuse surface is considered 1?

I am assuming if someone is using this function to input it to a neural net as roughness, they would want to say 1 instead of 0.0,  cause roughness=0.0 mean an extremely glossy surface.

include/mitsuba/render/bsdf.h
saeedhd96: Do you mean I should indent it one more tab or one less?
Isn't it currently in line with `diffuse_reflectance`?

src/bsdfs/twosided.cpp
saeedhd96: Would this be a concern of this PR? since such a refactoring is a different task than this, and should happen for other methods of this class, I would suggest doing that in another tab. Also, I am not too good at C++, and I cannot do such a refactoring without bad consequences :)

saeedhd96:
I can totally get the point of having it return two. roughness values for anisotropic. I can do that, but I guess that would be the behavior across all imlementations ... is that OK?
I am planning to use `mi.Vectorf2f() `for that, is that ok?

include/mitsuba/render/bsdf.h
njroussel: The `Mask active = true` argument should be aligned with the opening parenthesis of the previous line. Like in the other methods. Do you see what I mean?

include/mitsuba/render/bsdf.h
njroussel: I'm confused. 

Doesn't the default behaviour just return 1 currently? It doesn't use the BSDF evaluation at the normal direction.

src/bsdfs/twosided.cpp
njroussel: I'm ok with this. I'll refactor it with a lambda once it's merged.

src/render/bsdf.cpp
njroussel: Indeed, I agree.

njroussel:
I hadn't considered the anisotropic case.
It's not an issue, you can return two values for all implementations. `Vector2f` is good too :) 

(Sorry for the delay...)

njroussel:
Closing this as its outdated and I'm fairly certain it can be rewritten with `eval_attribute`.

PR #590, opened 2023-03-14
Remove obsolete tracking of Object class reference

dvicini:
I think this manual tracking of the class reference is now obsolete, since the separation of libcore and librender is gone. The double refcount decrease was always a bit suspicious here.


njroussel (approved):
LGTM, thank you!

PR #592, opened 2023-03-15
Parallelize load_dict routine

Speierers:
## Description

This PR aims to improve the performance of loading large Mitsuba scenes generated procedurally in Python by parallelizing the scene loading routine for Python dictionaries. The changes result in a significant speedup for large scenes (> 2x on my Macbook Pro M1).

The implementation of load_dict() in xml_v.cpp has been refactored to be more similar to the implementation of load_file in xml.cpp. First, the entire dictionary is traversed on a single thread to build a dependency graph and the corresponding Properties objects. Then, the loading of different instances is dispatched on multiple threads.

## Testing

The PR does not have any specific test cases since the default behavior is now set to `parallel=True`. This means that the changes have been implicitly tested across the test suite.

### DIFF
--- a/src/core/python/xml_v.cpp
+++ b/src/core/python/xml_v.cpp
@@ -5,17 +5,38 @@
 #include <mitsuba/core/spectrum.h>
 #include <mitsuba/core/transform.h>
 #include <mitsuba/python/python.h>
+#include <nanothread/nanothread.h>
 #include <map>
 
 using Caster = py::object(*)(mitsuba::Object *);
 extern Caster cast_object;
 
+struct DictInstance {
+    Properties props;
+    ref<Object> object = nullptr;
+    uint32_t scope;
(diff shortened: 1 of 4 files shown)


src/core/python/xml_v.cpp
```diff
@@ -334,17 +343,119 @@ std::vector<ref<Object>> load_dict(const std::string &dict_key,
...
+        // Construct the object with the parsed Properties
+        inst.object = PluginManager::instance()->create_object(props, class_);
+
+        if (!props.unqueried().empty())
+            Throw("Unreferenced property \"%s\" in plugin of type \"%s\"!", props.unqueried()[0], type);
+    };
+
+    // Top node always instantiated on the main thread
+    if (is_root) {
+        // Top node always instantiated on the main thread
```
njroussel, line 423: Remove one of the two comments.

Speierers:
Thanks @njroussel for the review 🚀

Issue #613, opened 2023-03-28
Bitmap box filter resampling artifacts


dvicini:
Hi,

I was playing around with the Bitmap's resampling method and tried it in a few different settings. It appears that for certain configurations of resampling with the `box` filter, we get some artifacts that are unexpected. The problem appears when using the box filter for upsampling. For example, upsampling from 128x128 to 200x200 is problematic. On the other hand, 128x128 to 256x256 works fine.

Here is an example result, where I also performed the same resampling using OpenCV, which seems to work.

<img width="928" alt="image" src="https://user-images.githubusercontent.com/5487532/228194884-7fd7540e-f208-4988-acc6-a25eb8a5b9c0.png">

```python
import matplotlib.pyplot as plt
import cv2
import mitsuba as mi
import numpy as np

mi.set_variant('scalar_rgb')

img = np.random.rand(128,128,3)

result = mi.Bitmap(img)
result = result.resample((200,200), mi.load_dict({'type': 'box'}))
result = np.array(result)

fig, ax = plt.subplots(1, 3, figsize=(12,4))
ax[0].imshow(img)
ax[1].imshow(result)
result_cv2 = cv2.resize(img, (200, 200), interpolation= cv2.INTER_AREA)
ax[2].imshow(result_cv2)

plt.show()


```

njroussel:
Hi @dvicini :smile: 

I can reproduce this. The good news is that we already have the necessary code path to detect whether or not we support a specific upsampling routine. The bad news is that it's currently checked in an assertion that is only compiled in `Debug` mode:
https://github.com/mitsuba-renderer/mitsuba3/blob/master/include/mitsuba/core/rfilter.h#L196-L198

I'm fine with making these into runtime exceptions in `Release`. I'd also welcome a PR if you actually need to handle such upsampling routine (I'm not so sure how this would work with a `box` filter).

dvicini:
Addressed in PR #625

PR #625, opened 2023-03-31
Fix upsampling using box filter

dvicini:
Fixes issue #613 

The problem is the usual problem of the box filter evaluation being numerically slightly unstable.

E.g., when upsampling an array from length 2 to 3, the middle pixel will evaluate a contribution from a lower level pixel exactly at 0.5 and therefore fall outside of the box filter support. 

```
|.....|.....| (input array)
|...|...|...| (output array)
      ^
      |
  problematic pixel
```

I added a special code path for box filtered upsampling. It's not super elegant, but a relatively minimal modification of the existing code. The alternative would be to add a completely separate box filter upsampling code path, which would be a bit cumbersome in this class. I also added a corresponding test.


njroussel (approved):
This looks good to me.

PR #666, opened 2023-04-15
Add Python bindings to access the underlying buffer of a MemoryStream

dvicini:
This adds an accessor function and python binding for the MemoryStream's underlying data buffer. I found this useful to be able to write Bitmap data to an arbitrary output stream in Python by going from Bitmap to MemoryStream to my custom output stream. Concretely, this is helpful to write Bitmap data to file systems that aren't supported by Mitsuba's file stream.


njroussel (approved):
Looks good to me :+1:

PR #667, opened 2023-04-15
Add support to write PLY files to a stream

dvicini:
Adds support to write PLY files to an arbitrary stream (e.g., `MemoryStream`). The design follows the `Bitmap` class.

### DIFF
--- a/src/render/python/shape_v.cpp
+++ b/src/render/python/shape_v.cpp
@@ -1,3 +1,4 @@
+#include <mitsuba/core/stream.h>
 #include <mitsuba/core/struct.h>
 #include <mitsuba/core/properties.h>
 #include <mitsuba/render/bsdf.h>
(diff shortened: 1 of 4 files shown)


src/render/python/shape_v.cpp
```diff
@@ -173,8 +174,12 @@ MI_PY_EXPORT(Shape) {
...
         .def_method(Mesh, has_vertex_normals)
         .def_method(Mesh, has_vertex_texcoords)
-        .def("write_ply", &Mesh::write_ply, "filename"_a,
-             "Export mesh as a binary PLY file")
+        .def("write_ply",
+             py::overload_cast<const std::string &>(&Mesh::write_ply, py::const_),
+             "filename"_a, "Export mesh as a binary PLY file")
+        .def("write_ply",
+             py::overload_cast<Stream *>(&Mesh::write_ply, py::const_),
+             "stream"_a, "Write mesh as a binary PLY to a stream")
```
njroussel, line 182: Any specific reason why you're not using the `D()` macro here?

njroussel (commented):
One tiny comment. The rest is great.

Issue #655, opened 2023-04-12
PRB memory usage grows with number of SPP


Speierers:
While profiling inverse rendering optimization using the `prb` integrator, I realized that the memory usage of the integrator grows as I increase the number of samples per pixel (or the render resolution). My understanding was that such integrator works in a megakernel fashion, hence thread rendering state should stay in registers and only the output image / gradients would be stored in VRAM.

Looking at `common.py`, I noticed that we are scheduling the `state_out` variable in between the different phases of the PRB algorithm (see the following [line](https://github.com/mitsuba-renderer/mitsuba3/blob/master/src/python/python/ad/integrators/common.py#L937)). If not scheduled, I observe the expected behaviour (e.g. no scaling in memory usage).

Would it make sense to **not** schedule this variable, hence "fusing" the two phases in a single kernel launch but saving us from extensive memory usage? I suppose that this would also duplicate some of the computation, as a first kernel will be launched when developing the film (phase 1), and a second kernel will be launched for phase 2, re-computing the `state_out` values in this second kernel. Further reordering in this method will likely be necessary to completely fuse the two phases into a single `ray_gen` kernel. Any other option?

The same seems to be true for the `render_forward()` method, where the `state_out` variable is explicitly evaluated.

njroussel:
Hi @Speierers  :smile: 

Let's write this out explicitly to avoid confusion (considering only the backward case):

During the `render_backward()` we currently have 3 kernels:
1) Does a primal rendering (first phase). This kernel's size is equal to the total number of samples.
2) Develops the film of the primal rendering (first phase). This kernel's size is equal to  `#pixels * #channels`.
3) Does the backward mode rendering (second phase). This kernel's size is equal to the total number of samples.

If you do not schedule `state_out` you'll still get 3 kernels with the difference that 3) practically re-computes 1). This is specifically true for `prb` as `state_out` matches `L`.
I guess my confusion comes from your use of "fusing". We will **always** have 3 kernels, right? The first two kernels will always be split apart because of their different sizes. The difference here is what kernel 3) computes and what input variables it will take.

I understand your initial point and agree that we should at least offer the option of this tradeoff: virtually rendering once more vs. using more VRAM memory.

njroussel:
I just don't see how you could compute 3) (or 4) for that matter) without knowing 2). You need the differential radiance `δL` to compute 3) which I believe you can only get from 2). 
[...]

PR #672, opened 2023-04-17
Fix memory usage of PRB integrators

Speierers:
## Description

This PR reorganizes the kernels generated by `RBIntegrator.render_backward()` in order to avoid the need to store intermediate rendering state to memory (e.g. `state_out`).  This issue was originally raised in #655.

With the patch, the integrator first computes the adjoint radiance by differentiating the sample splatting and weight division. This doesn't require any information from the 1st phase of PRB, hence can be done independently. This will generate two kernels: one for the ray generation, and one for the splatting. Then the integrator generates a fused kernel that computes phase 1 and 2 of PRB. Here only the ray generation operations are duplicated, which shouldn't a real impact on performances. Finally, for reparameterized integrators (e.g. `prb_reparam`), the radiance computed in phase 1 (e.g. `L`) is backpropagated through the film splatting and weight division to compute the gradients of the sample positions. Compared to the original implementation, this will introduce an extra kernel that computes the splatting a second time in the case of reparameterized integrator.

This PR also adds `Film::clear()` that clears the content of a film (e.g. after calling `put_block()`), needed for this patch.

## Example

To test this patch, here are some statistics and information about the kernels launched in the `render_backward()` method when differentiating the color of the red wall of the Cornell Box scene with 4096 samples per pixels.

### Before the patch

**Kernels**:
1. PHASE 1: (n=268435456, in=54, out=3, se=100, ops=1821)
2. ADJOINT RADIANCE: (n=196608, in=3, out=0, se=2, ops=35)
3. PHASE 2: (n=268435456, in=63, out=0, se=1, ops=2118)

**Memory usage**:
2.042 MiB/3.003 GiB used (peak: 3.003 GiB)

### After the patch

**Kernels**:
1. ADJOINT RADIANCE (ray gen): (n=268435456, in=3, out=0, se=100, ops=665):
2. ADJOINT RADIANCE (splat): (n=196608, in=3, out=0, se=2, ops=35):
3. PHASE 1 + 2: (n=268435456, in=80, out=0, se=1, ops=2775)

**Memory usage**:
1.042 MiB/3.047 MiB used (peak: 3.047 MiB)

Not only the memory usage is drastically reduced as it is now independent of the SPP, but the total number of operations over the 3 kernels is smaller after the patch.

## Testing

This patch should be covered by the tests in `test_ad_integrators.py`.

### DIFF
--- a/src/python/python/ad/integrators/common.py
+++ b/src/python/python/ad/integrators/common.py
@@ -63,6 +63,8 @@ class ADIntegrator(mi.CppADIntegrator):
 
         if isinstance(sensor, int):
             sensor = scene.sensors()[sensor]
+        
+        film = sensor.film()
 
         # Disable derivatives in all of the following
         with dr.suspend_grad():
@@ -91,16 +93,16 @@ class ADIntegrator(mi.CppADIntegrator):
             )
 
             # Prepare an ImageBlock as specified by the film
-            block = sensor.film().create_block()
+            block = film.create_block()
 
             # Only use the coalescing feature when rendering enough samples
             block.set_coalesce(block.coalesce() and spp >= 4)
 
             # Accumulate into the image block
             alpha = dr.select(valid, mi.Float(1), mi.Float(0))
-            if mi.has_flag(sensor.film().flags(), mi.FilmFlags.Special):
-                aovs = sensor.film().prepare_sample(L * weight, ray.wavelengths,
-                                                    block.channel_count(), alpha=alpha)
+            if mi.has_flag(film.flags(), mi.FilmFlags.Special):
+                aovs = film.prepare_sample(L * weight, ray.wavelengths,
+                                           block.channel_count(), alpha=alpha)
                 block.put(pos, aovs)
                 del aovs
             else:
@@ -111,8 +113,8 @@ class ADIntegrator(mi.CppADIntegrator):
             gc.collect()
 
             # Perform the weight division and return an image tensor
-            sensor.film().put_block(block)
-            self.primal_image = sensor.film().develop()
+            film.put_block(block)
+            self.primal_image = film.develop()
 
             return self.primal_image
 
@@ -172,11 +174,11 @@ class ADIntegrator(mi.CppADIntegrator):
                 #   Σ (fi Li det)
                 #  ---------------
                 #   Σ (fi det)
-                if (dr.all(mi.has_flag(sensor.film().flags(), mi.FilmFlags.Special))):
-                    aovs = sensor.film().prepare_sample(L * weight * det, ray.wavelengths,
-                                                        block.channel_count(),
-                                                        weight=det,
-                                                        alpha=dr.select(valid, mi.Float(1), mi.Float(0)))
+                if (dr.all(mi.has_flag(film.flags(), mi.FilmFlags.Special))):
+                    aovs = film.prepare_sample(L * weight * det, ray.wavelengths,
+                                               block.channel_count(),
+                                               weight=det,
+                                               alpha=dr.select(valid, mi.Float(1), mi.Float(0)))
                     block.put(pos, aovs)
                     del aovs
                 else:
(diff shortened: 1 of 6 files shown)


dvicini:
Hi Sebastien,

I only looked at the non-reparameterized part and that one looks good to me. Nice work! 

My only comment would be that in principle we could have a fast path for the box filter there: in case of a box filter, we should just directly look up the pixel in  `grad_in` and divide by SPP (i.e., no need for actually splatting and calling backward). I think that's what we did in the original implementation in the PRB paper, and the handling of non-box filters was only added when working on reparameterizations.

But maybe this fast path can be implemented later, as it's orthogonal from the proposed changes.

Speierers:
IIRC with a box filter most of the computation is opt out in the imageblock/film and it will almost fall back to a simple gather. We just add the cost of jitting to build the AD graph.

njroussel (approved):
This is great!

I see that you haven't changed the `render_forward()` method. The same trick should also just apply there. Is this something you've already done? I can do it myself otherwise.

njroussel:
Fixes #672

Speierers:
> This is great!
> 
> I see that you haven't changed the `render_forward()` method. The same trick should also just apply there. Is this something you've already done? I can do it myself otherwise.

No, I haven't looked at the `render_forward()` method. Your help would be greatly appreciated as I won't be able to work on this in the coming week. It shouldn't be simpler that the `render_backward()` method as the 3rd phase doesn't require `dL` to be computed.

Speierers:
My latest commit includes similar changes to the `render_forward()` method.

## Example

Running the same example as above, this time propagation the gradients forward.

### Before the patch

**Kernels:**

(n=268435456, in=53, out=3, ops=1220)
(n=268435456, in=61, out=0, se=1, ops=1485)
(n=196608, in=2, out=1, ops=23)
(n=1, in=1, out=1, ops=4)

**Memory usage:**

2.042 MiB/3.003 GiB used (peak: 3.003 GiB)

### After the patch

**Kernels:**

(n=268435456, in=78, out=0, se=1, ops=2146)
(n=196608, in=2, out=1, ops=23)
(n=1, in=1, out=1, ops=4)

**Memory usage:**

2.042 MiB/3.047 MiB used (peak: 3.047 MiB)

wjakob:
Hi @Speierers  -- do you have an intuition about why the peak memory reduction does not occur in the forward version?

wjakob:
Ah wow. `GiB` vs `MiB` :-) . I only looked at the number.

wjakob:
One request from me: the ``Differentiate sample splatting and weight division`` step now appears in a few places and seems somewhat repetitive. Do you think it would be possible to factor out a generic version to reduce the amount of repitition?

src/python/python/ad/integrators/common.py
```diff
@@ -477,6 +462,30 @@ def prepare(self,
 
         return sampler, spp
 
+    def _splat_to_block(block: mi.ImageBlock,
```
wjakob, line 465: It's not really a simple splat (aka atomic-add), right? (given the weight division).

This might be more difficult, but can the whole "differentiate a write to get a read" be factored out? (That's what initially had in mind, and it would apply specifically to the reverse mode parts).

wjakob:
(I will let Nicolas do the final merge).

src/python/python/ad/integrators/common.py
Speierers: I think we are talking about two seperate things. 

1. I factored out the block splatting (HDR vs Spec film) logic in `_splat_to_block`. This is not about differentiation, e.g. it is used in the primal rendering method as well.
2. In `render_backward()` I factored out the backpropagation code that pushes the gradient image through the splatting + weight division. You can find this code in the local function `splatting_and_backward_gradient_image`, which is used twice in `render_backward()`.

Is this more clear, or you meant that something else needs to be factored out?

njroussel:
I'm happy with the current iteration, thank you for the refactorings @Speierers!

In the future, we can revisit the differentiation steps of the `ImageBlock` and `Film`, it still believe that we can simplify it.

PR #678, opened 2023-04-18
[✨ feature] Emissive Volume Rendering

Microno95:
## Description

This PR adds emissive event sampling in the null-scattering volpathmis integrator for sampling the VRE and MIS for emission. It supersedes the previous PR https://github.com/mitsuba-renderer/mitsuba3/pull/40.

Adds emissive volume support by adding a volume emitter. The interface adds a `volumelight` emitter to a given shape that is then sampled as part of the medium sampling routines in volpath/volpathmis/prbvolpath.

The current implementation modifies none of the interfaces of endpoint and thus does not work with MIS as-of-yet. I am unsure of the best way to implement this in terms of modifying the interface. The cleanest solution seems to be that the endpoint sampling methods take an extra float to sample a 3d point, but whether this should be implemented by changing the existing Point2f sample to Point3f or adding an additional argument is unclear to me.

The current implementation is feature complete except for NEE for volume emitters, I have added and tested the python interfaces by bringing the prbvolpath implementation to parity with volpath.

## Checklist

- [x] My changes generate no new warnings ~~- generates warnings about unused parameters in `volumelight`, these methods will be implemented and the parameters used.~~ There are no longer any new warnings.
- [x] My code also compiles for `cuda_*` and `llvm_*` variants. I have compiled every single variant and run all the tests, and everything passes except for the tutorial notebooks, but these throw an unicode char encoding error so I think it's unrelated. I will check with a fresh master copy if the same error occurs, might be a windows only issue.
- [x] I have added tests that prove my fix is effective or that my feature works - ~~this is in-progress, have added emissive volume scenes and tests that I will make a PR for in the [mitsuba-renderer/mitsuba-data](https://github.com/mitsuba-renderer/mitsuba-data) repo.~~ Created the following PR to merge emissive volume tests: mitsuba-renderer/mitsuba-data#17

## Notes
* The implementation uses non-analogue probabilities as described in [Monte Carlo Methods for Volumetric Light Transport Simulation](https://cs.dartmouth.edu/wjarosz/publications/novak18monte.pdf). This leads to more consistent results when a given radiance channel is zero and better convergence in all the tests I've done. 
* Both `volpath` and `prbvolpath` have implementations of the analogue, maximum and mean event probability calculations for testing. The issue of inconsistency only appears with analogue probabilities in non-scalar modes, and only for volpathmis and prbvolpath.
* Sorry for the delay in the implementation! The move to Mitsuba 3 meant I had to rewrite a lot of the code, but I managed to be significantly more thorough this time around so everything compiles and all the tests pass.
* I've attached an example render as well! 
homogeneous-emissive-cuda


Microno95:
I've identified the issue with the failing tutorial notebook, it's specifically related to the prb integrator and my incorrect implementation of accumulating radiance. I'm not quite sure what the correct implementation is as my understanding of it is that the radiance accumulated from the medium should be treated similarly to intersection with emitters, but this seems to break the differentiability of the rest of the medium parameters.

dvicini:
Nice work! Sorry I somehow missed this new PR.

Just a high level comment for now: This adds quite a lot of code complexity - would it make sense to have special integrators for emissive media? It's somewhat of a special case and the existing volpath integrators are already hard to parse (e.g., for someone coming from using NeRF...)

Microno95:
I have significantly simplified the sampling of emission and reduced the complexity of the integrators in my latest code that I will push soon.

Generally, I think it does make sense to move these to the medium class as the different medium event sampling modes are better suited for different media. e.g. if a medium has 0 scattering for, say the green channel, the analogue probabilities actually underestimate the number of scatters when using hero wavelength sampling and lead to significant fireflies.

Turns out you can sample radiance at every medium interaction (at least when not using NEE for medium emitters, that's a separate case to be worked on), and this has simplified the inclusion of radiative media effects to a single additional line in both `volpath` and `volpathmis` integrators and only 4 additional lines in `prbvolpath` (a few more for the derivative tracking).

I have held off on pushing any updates as I was waiting on how I should go about updating the endpoint/emitter interface to support a third random sample.

dvicini:
I see, thanks for clarifying.

I don't think adding an extra sample to the endpoint interface is a big deal. 

How do you deal with sampling the volume of complex shapes? Do you simply sample the bounding box and fail the sample if it falls outside the shape? I think currently it's still tricky to use an arbitrary number of random samples within a virtual function call (e.g., to do rejection sampling)

Microno95:
Does it make sense to add the 3rd volume sample by changing the `Point2f` sample to a `Point3f` sample? It seems the least cluttered way and is the most logical extension. But happy to also make it an additional parameter in the interface.

I have not yet decided on the best way to sample triangle mesh-based shapes. My two ideas were to either sample within the bounding box and fail the sample if it falls outside as you said, or to build a discrete distrubution using a simple tetrahedral decomposition of the mesh. 

The upside of using a tetrahedral representation (or really any discretisation of the volume) is that it'll be faster to sample and easier to check if a sample falls inside/outside the shape.

The alternative is iterating over every triangle to calculate a winding number or using ray intersections (calculating the number of crossings). The issue with ray intersections is in regards to choosing a direction for testing and I've found that it can be problematic depending on the mesh and winding number calculation scales linearly with the number of triangles.

dvicini:
I think changing the sample to be Point3f isn't a big deal, better than adding another parameter.

I don't think adding a tetrahedral representation makes sense (from an engineering point of view). As far as I know - correct me on that - computing a tetrahedralization is far from trivial and not something that can easily be implemented in. a standalone helper function. 

I would suggest to check if a point is inside the mesh by simply tracing a ray in a fixed (or randomized) direction and checking if the triangle we hit is facing away from the ray direction. This should work for watertight meshes, and that either way is the assumption for any of the volumetric integrators to work correctly.

Microno95:
That makes sense! I will work on those and push an update asap.

Microno95:
Hey @dvicini, I am in the process of implementing the interface for Next Event Estimation of volume emitters, but I'm struggling with estimating the pdf of a direction sampled in the emitter, specifically implementing the `sample_direction` and `pdf_direction` methods.

The issue I'm having is that while I know the pdf of sampling a given point in the medium, the pdf of a direction is with respect to the solid angle whereas the pdf of sampling a point is with respect to the volume. I'm not sure how to make these two commensurate.

I know the following relation should hold

$$
\int_\Omega p(\vec{\omega}) \mathrm{d\Omega} = \int_V p(x) \mathrm{dV}
$$

i.e. the integral of the solid angle pdf over all solid angles should be equal to the integral of the volume pdf over all volumes (and both should be equal to 1 for a normalised pdf.). Knowing that, I can rewrite the equation as follows.

$$
\int_\Omega p(\vec{\omega}) \mathrm{d\Omega} = \int_\Omega\int_r p(x) r^2\mathrm{dr}\mathrm{d\Omega}
$$

And (this is the part I'm unsure about), this should imply that

$$
p(\vec{\omega}) \mathrm{d\Omega} =\int_r p(x) r^2\mathrm{dr}\mathrm{d\Omega}
$$

If all that is correct, then the pdf of sampling a given direction should be straightforward to compute based on the line segments that pass through the medium - easily found via ray intersections. I'm not sure how this is to be treated in the case of medium vs bsdf interactions given that $\mathrm{d\Omega}$ has two different domains, and how to handle the case where a given point is inside the medium.

dvicini:
The Jacobian to change between solid angle and volume sampling is indeed $r^2$. This is exactly the geometry term in a volume.

I am not sure I fully understand you question. if this is primarily about MIS, you will just want to make sure that both PDFs are in the same measure. 

The conceptually simplest way is to integrate the volume PDF over the ray segment, but that is costly and potentially inaccurate (e.g. can be done using ray marching). It's good to keep in mind that MIS weights don't have to use exact PDFs, they just need to be computed consistently to ensure unbiasedness.

Another option could be to compute the MIS in volume measure, I think that would forego the need to integrate over segments (potentially at an increase in noise? but with cheaper per-sample cost). In that case, you would just have to convert the solid angle PDF to volume PDF using the geometry term.

Microno95:
Right, I think I understand better now. I was thinking I'd need to convert the volume PDF to a solid angle PDF in order to do MIS with the volume emitter, but you're saying I can do it the other way around (which makes perfect sense).

In terms of converting the solid angle PDF to a volume PDF, would it be correct to divide $p(\vec{\omega})$ by $r^2$ or does it require something else? i.e. the MIS is computed via $\mathrm{MIS}(r^{-2}p(\vec\omega), p(\vec x))$

dvicini:
Yes, this sounds good to me. 

You will need to make sure that the MIS is consistent with the evaluation of emission after direction sampling. I.e., the simplest method would again be to evaluate the emission along the entire ray segment.

Microno95:
It seems that simply converting the solid angle PDF to a volume PDF does not work, or perhaps I'm doing something wrong. I've verified that integrating the volume PDF over the ray length works, but it is slower (by a lot, around 20x-40x due to the ray casts required) and raises OptiX warnings in CUDA modes.

At first I thought it would be a simple change of variables using the differential element $$r^{-2}\frac{\mathrm{dV}}{\mathrm{d\Omega}}$$, but this gives very different results compared to integrating over the ray length and is inconsistent with the previous results obtained without NEE.

Is it perhaps to do with the fact that $r^{-2}p(\vec{\omega})dV$ is a non-normalisable PDF? i.e. the following integral does not converge

$$
\int_V r^{-2} p(\vec{\omega}) \mathrm{dV} = \int_V  \left(r^{-2}p(\vec{\omega})\right) \mathrm{r^2\sin\theta dr d\theta d\phi} = \int_V  p(\vec{\omega})\mathrm{\sin\theta dr d\theta d\phi}
$$

Microno95:
The current implementation correctly tracks the ray entry and exit points in a given shape and calculates the exact directional pdf. In `llvm_*` modes this is quite slow compared to not using NEE by around 5x while in `cuda_*` modes the discrepancy is much less at around 1.5x slower with a noticeable reduction in the noise for the same sample count.

There is one big issue in that in `llvm_*` modes, the ray entry and exit points lead to different results compared to `cuda_*` modes, most likely due to the different handling of ray intersections by Embree vs OptiX and thus the current implementation only gives correct results for `cuda_*` modes.

I am working on improving and fixing the implementation, but I would welcome a second set of eyes over the code to see if there's anything I'm missing in how I'm accumulating the pdf.

Microno95:
Looking further into, it seems that `cuda_*` modes gave the correct results when the ray intersection routines were disabled giving a pdf of 0 - this was due to how I initialized the nested volume scene allowing me to do ray intersections. Ultimately this gave the correct result as it effectively disabled NEE.

I've attached two renders, with and without NEE (512x512@4096spp), that show the discrepancy between the NEE and non-NEE results. This is of a `sphere` shape with a purely absorbing medium with a radiance of 1 and an extinction coefficient of 1 at the origin with a radius of 1.

The discrepancy definitely has to do with how the PDF is computed, but I'm not sure what's wrong as I've verified that the calculations are correct - by computing the analytic solution to the directional PDF and comparing it to the ray traced solution for the `sphere` shape as per our earlier discussions. From the emission gathered at the floor and back wall, it looks like the NEE version is underestimating the illumination from the medium - perhaps there is a term that has to be accounted for.

Without NEE:
emitting_sphere_REF

With NEE:
emitting_sphere_NEE

I've also attached the .exr files themselves for better numerical comparison in TEV.
[emitting_sphere_renders.zip](https://github.com/mitsuba-renderer/mitsuba3/files/11800249/emitting_sphere_renders.zip)

dvicini:
Hmm, really hard to say what's wrong just from these two images. I would say that for a homogeneous sphere of emissive volume it should really be possible to get a perfect match, as it doesn't have any of the complexity of complex shapes or delta tracking for heterogeneous media. My only suggestion would be to further reduce the complexity: 1) only render direct illumination, no scattering and 2) completely disable absorption. I.e., can you match the total emission from an emissive sphere volume that does not absorb any light?

Microno95:
With a homogeneous sphere where $\sigma_t=\sigma_s=0$ and $\sigma_n=1$ (in order to gather emission from the medium), the renders still diverge for any ray max depth $>0$. For a max ray depth of 1, so just purely illumination from emitter intersections, the two renders agree perfectly. For a max ray depth of 2 or greater, they diverge with the NEE renders looking similarly dimmer as before.

In the literature, the closest I've seen to the implementation of the solid angle PDF is in Equation 15 of [Line Integration for Rendering Heterogeneous Emissive Volumes](https://onlinelibrary.wiley.com/doi/full/10.1111/cgf.13228) where they estimate the solid angle PDF by integrating the volume PDF along the ray direction with the appropriate jacobian.

Perhaps there is additional consideration in how NEE estimates of illumination are combined with Unidirectional estimates for volume emitters.

Microno95:
I've looked further into it, and I think there's actually a bug in how `volpath` handles NEE. I've created a case where a surface emitter with a diffuse reflectance of 0 gives incorrect results with `volpath` when emitter sampling is enabled, regardless of whether or not there are media in the scene at all.

Here's the reference render using the `Path` integrator:
ref_path_render

And here's the render from `VolPath`:
incorrect_volpath_render

I've also included the renders from `volpathmis` and `prbvolpath` as well.

`volpathmis`:
incorrect_volpathmis_render

`prbvolpath`:
incorrect_prbvolpath_render

I've tested it and the issue exists in the `master` branch so I am sure it is unrelated to my changes to the code. I've attached the scene here:
[minimal_example.zip](https://github.com/mitsuba-renderer/mitsuba3/files/11823814/minimal_example.zip)

dvicini:
Oh, that's a good catch. Not sure what's wrong here, but maybe something related to the ShadowEpsilon or so. I will take a look, would be good to fix

dvicini:
I think the problem is a bit more subtle. By adding the epsilon to the check for the remaining distance, you are essentially doubling the shadow epsilon. The epsilon is already included in the computation of `remaining_dist`. The better solution is to replace the `spawn_ray` function call with a call to `spawn_ray_to`, which has a better epsilon calculation. Then, we can also replace the use of `ds.dist` with the `maxt` set within `spawn_ray_to`. This is more consistent with what the path integrator does. I will make a PR

Microno95:
In order to simplify some of testing, I separated out emitter sampling and unidirectional sampling in `volpath`. Separately, the two methods give near identical results for an analytic shape like a sphere even in the presence of absorption and volumetric scattering. 

Here are two (low-res) renders to show the results:
| Unidirectional Sampling | Emitter Sampling |
| ----------- | ----------- |
| drawing | drawing |

The convergence for emitter sampling is atrocious, it requires an order of magnitude more samples for this case than unidirectional sampling - most likely due to the fact that I can control how densely unidirectional sampling accumulates emission via $\sigma_n$, but emitter sampling is indifferent to it. Either way, I can at least show the implementation of the sampling procedure is correct.

There are now two issues to resolve:
1) Rejection-like sampling of the mesh isn't working and I'm not exactly sure why. It seems that setting the PDF to $0$ when outside the shape and $V^{-1}$ when inside gives incorrect results - it significantly underestimates the emission compared to unidirectional sampling.
2) MIS between the two sampling procedures is problematic as the PDF of a given direction is not compatible with the fact that I'm accumulating radiance from a single point rather than the emission along the ray segment passing through the medium in that direction. For this, I think a possible solution is to use the same direction sampling procedure, but instead of accumulating the radiance from a point, accumulate the radiance along the ray in `sample_emitter`. I will try that and see how it goes...

Microno95:
I apologise for turning this pull request comment thread into a stream of thought type log, but it has helped me work through the code and I'm hoping that it provides some insight into the implementation when it's finalised.

I have a version of the code that works, it uses emitter sampling to sample the emitting sphere media and estimates the radiance along the ray through the medium, and can MIS the results with unidirectional sampling. This only works in `volpath` at the moment, but it works:

| Unidirectional Sampling | Emitter Sampling | Multiple Importance Sampling |
| ----------- | ----------- | ----------- |
| drawing | drawing | drawing |

dvicini:
Nice, looks good! FYI, the PR fixing the epsilon issues has been merged in the meantime.

Microno95:
Hey @dvicini, 

So I've got a solution implemented for NEE for both the `volpath` and `prbvolpath` integrators and they agree with each other in terms of results and with the no-NEE integrators. I'm still working on the `volpathmis` solution, but how emitter sampling is done is somewhat more complicated as the radiance along the ray passing through the medium is estimated by the medium stepping routine in the emitter sampling code.

I'll hopefully have that sorted in the next day or two. What I wanted to ask about is how to correctly implement the adjoint for the Path Replay Backpropagation version of the `volpath` integrator. I've an implementation in the repository, but I'm not sure it is entirely correct and I'm not sure how best to check it - would using the `inverse_rendering/volume_optimization.ipynb` notebook be a good start by trying to do the optimisation with an emissive medium instead of an absorbing/scattering one or is there something more basic that occurs to you?

dvicini:
Hi,

That sounds great! Yes, I think the volume_optimization notebook should be a good starting point to work with the PRB version of the integrator. You should also take a look at the `test_ad_integrators.py` file to see how we run finite-difference validation for our differentiable integrators. For the existing PRB integrators, we implemented both a forward and reverse-mode version. The forward mode version can be used to output "gradient images" which allow to more easily visually compare to a finite difference gradient (for 1 single parameter in this case).

I am not sure i understand the question about delta/ratio tracking. We use delta tracking throughout the medium and ratio tracking to evaluate the transmittance term during NEE.

dvicini:
For the loop recording bug: Did you ensure all variables that are modified are listed in the Loop construction? That's usually the source of these kind of crashes. 

For the FD tests: I am not sure what is currently being tested for the PRB volpath integrator, but I just double checked and see that you are right that it doesn't support forward mode. I got confused because there is a superfluous `dr.forward_to(Lo)` in the integrator code, which will never be reached.

Microno95:
I did ensure that, perhaps there is a side-effect of some variable I'm accessing or creating with a method call, but would those cause issues? One thing that comes to mind is how I get the radiance from the underlying emitter plugin which is by creating a SurfaceInteraction from a MediumInteraction. Could this cause issues during loop recording?

dvicini:
I would just do the usual, commenting out parts of the code until it works and narrowing down where the issue comes from. Unless the issue is obvious, I found this to be the best way to quickly localize what is going on.

Creating a surface interaction from a medium interaction might result in fields that are uninitialized? In that case, maybe do something like 
```
si = dr.zero(SurfaceInteraction3f)
si.p = mi.p
si.t = mi. t
...
```

Microno95:
I tried everything and it turns out the issue was related to an accumulation variable I was using. More specifically, I was accumulating the sampled medium radiance in a variable `medium_contribution` every iteration of the PRBVolpath NEE loop. I had added this to the series of loop variables as it's a relevant quantity, but it turns out that doing so caused the Loop initialisation to fail and thus led to a full Python Fatal Error on the backward pass.

It's a very strange thing, not really sure why it does that, but having removed the variable from the loop variables the integrator seems to work and gives correct results.

Microno95:
Right, so I had a reply yesterday suggesting things were fixed and I found that they weren't. It finally makes sense what's going wrong and it's due to the fact that the variable I need differentiated is also being accumulated into in the loop. That doesn't work because differentiating loops is not supported. It also explains why it works with loop recording disabled.

I now have to figure out the exact form to pass to `dr.backward` to get the correct gradients on each iteration of the loop...

The `Fatal Python Error: Aborted` is due to the issue in #928, as, on the traceback, python tried to get the string representation of a placeholder variable - bypassed this and found it was actually because I was trying to differentiate the loop.

Microno95:
This latest commit includes a forked version of the resources needed for the tests (for which I've created mitsuba-renderer/mitsuba-data#17). I will be further adding AD tests for heterogeneous media for the `volpath` integrators, but for now this should cover almost all homogeneous cases and accurately test the efficacy of the Next Event Estimation that's been implemented. The integrators are fully self-consistent in that not only do they produce results matching each other, but also have gradients matching finite differencing.

I will further extend the implementation to `volpathmis` and write an additional integrator, `volpathdirect` that will implement the very basic delta-tracking algorithm with no NEE in order to verify that the correctness of the integrators - once that's done, I will consider this implementation complete.

Microno95:
Hey @dvicini,

So I got a few more tests set up for `prbvolpath` as I want to be sure that there are no issues before moving on. I was wondering what is the correctness of gradients computed via forward/backward differentiation of the `volpath` integrator?

I ask because I also have similar errors arising for `prbvolpath` and I wanted to know whether I should take the FD as ground truth or `volpath` gradients. There are a few cases where the `volpath` integrator gradients are very wrong (eg. zero where they're not supposed to be or a factor of 2 larger than FD) especially when backward differentiated. I did try varying the step size for FD and I'm using an 8th order 9 point finite difference stencil as the simple central differencing was too inaccurate (much more step size sensitive).

One of the `prbvolpath` issues was related to the fact that I have to sample the radiance with a detached `MediumInteraction` _and_ do `dr.backward` in a specific order otherwise the gradient is zero neither of which were initially obvious to me.

dvicini:
Hi,

I don't think we expect anyone to use `volpath` for differentiable rendering. The used delta tracking algorithm is an extremely poor fit for naive AD. So I am not surprised that `volpath` wouldn't give correct gradients, it's barely been tested for AD.

For FD, are you ensuring that you are using the same seed for all the evaluations? Otherwise, FD will be too noisy to be useful or require an absurd number of samples to converge.

`prbvolpath` on the other hand should be correct, except for the known limitation of density derivatives in vacuum regions (which was addressed by a follow up paper).

Microno95:
Ah ok, that explains why volpath is only mostly correct - on point except for scattering emissive media and density differentiation with a constant emitter.

As for the FD, I am using the exact same seed so I know that it's not noisy and I've additionally checked that it's not an issue with underflow due to too small a step size by checking the convergence of the FD as a function of step size.

The case where `prbvolpath` is incorrect is one where the optical density is differentiated and the medium is emissive, scattering and optically dense. A contrived case for sure, but it seemed appropriate to test the integrator in that scenario nonetheless.

Microno95:
Hey @dvicini,

I don't know what happened, I thought the runner would be killed for going overtime, but the linux build test runners are hanging and I didn't catch that until now. I have fixed the cause of the hang, but they're somehow 55 hours overtime, sorry...

In other news, I've fixed PRBVolpath as much as I could and even added forward-mode differentiation support in the process (which I've tested against all the integrator AD tests).

Microno95:
Hey @dvicini,

I don't know what happened, I thought the runner would be killed for going overtime, but the linux build test runners are hanging and I didn't catch that until now. I have fixed the cause of the hang, but they're somehow 55 hours overtime, sorry...

In other news, I've fixed PRBVolpath as much as I could and even added forward-mode differentiation support in the process (which I've tested against all the integrator AD tests).

Microno95:
I looked for it but couldn't find it most likely because I can only view the runners as a guest. Thought maybe you would have access or know someone who has that authority 😅 

You mean in reference to the modifications to `prbvolpath` right? The answer is no, I monkey-patched the old `prbvolpath` to test if it showed the same issue, and it did. The following is the test case that is failing.

```python
class MediumConstantEmitterAlbedoConfig(ConfigBase):
    def __init__(self) -> None:
        super().__init__()
        self.key = 'sphere.interior_medium.albedo.value.value'
        self.scene_dict = {
            "type":"scene",
            "plane":{
                "type":"rectangle",
                "bsdf":{
                    "type":"diffuse"
                }
            },
            "sphere":{
                "type":"sphere",
                "bsdf":{
                    "type":"null"
                },
                "to_world": T.scale(0.25),
                "interior":{
                    "type":"homogeneous",
                    "albedo":0.5,
                    "sigma_t":1.0
                }
            },
            "light":{
                "type":"constant"
            }
        }
        self.spp = 4096
        self.res = 16
        self.error_max_threshold = 1.0
        self.error_mean_threshold = 0.1
        self.error_mean_threshold_bwd = 0.125
        self.ref_fd_epsilon = 1e-2
```

Regarding the complexity of `prbvolpath`, I'm happy to separate the two implementations into an emissive and non-emissive one, but it isn't the final version as I haven't evaluated if there are any code redundancies or areas of simplification. I'm also happy to add thorough documentation in the form of comments to explain what each section is doing and why.

marcoromelli:
This looks like a great work! I'm not expert enough to judge myself so I directly ask you:

- Is this going to take into account polarization?
- When do you think this will be merged?

Microno95:
> This looks like a great work! I'm not expert enough to judge myself so I directly ask you:
> 
>     * Is this going to take into account polarization?
> 
>     * When do you think this will be merged?

Hi @marcoromelli, 

Thanks! To answer your question:
* The current implementation does not account for polarisation as a) I'm not an expert on polarisation in volumetric media and b) this implementation is meant to be a robust but minimum implementation of emissive media with NEE.
* I cannot give a timeline for the merge, I'm mostly waiting to know whether or not I should split the integrator implementation into two separate ones. Additionally, some tests crash on different platforms which I'm struggling to debug.

johnson-c:
Is there a timeline on merging this request? This is starting to be adopted more broadly in the fusion research community, it would be great to get this functionality into the main branch.

Microno95:
Alright so that took longer than I had hoped, but after a week of extensive testing and updating of the code, I have now brought everything up-to-date with the latest Mitsuba master branch (54dda7b7f3ac412d43e667ce07668494de8292c3) as of today. There are some TODOs still on my list and I've noted some of the caveats of the implementation below. 

Additional features that this update brings are:
  * `PRBVolpath` now fully supports forward-mode and reverse-mode differentiation of all scene parameters (except shape deformations). Mostly correct, slight bias when differentiating medium density $\sigma_t$ and phase function $g$ with the gradients having the correct sign but not the correct magnitude. This existed in the original `PRBVolpath` implementation so I marked them as expected failures.
  * Full testing of many volumetric cases in primal mode for `Volpath` and `PRBVolpath`, and forward- and reverse-mode for `PRBVolpath` in `test_ad_integrators.py` to ensure robust volumetric gradients.
  * `VolpathMIS` now does full MIS of the illumination from media which should further reduce chroma noise when there is high variation between spectral channels of the medium. This needs a bit more testing which I have on my TODO list by comparing to PBRT-V4, but it passes every test I've thrown at it including the pre-existing ones so I don't think I've broken anything nor introduced further bugs.

TODOs:
- [ ] Test mesh sampling and directional PDFs are correct
- [ ] Validate `heterogeneous` media results
- [ ] Update reference images in `resources` submodule
- [ ] Validate further with PBRT-V4 as reference. Especially

Caveats:
  * PRBVolpath gives incorrect gradients (correct sign, incorrect magnitude) for the medium density `sigma_t` and the phase function `phase`, but this exists in the original implementation as well.
  * The efficiency of mesh sampling is proportional to the fraction of the volume of the bounding box occupied by the mesh itself. I tried multiple methods of sampling the mesh, and found sampling the bounding box to be the most robust and simplest implementation.

Microno95:
Several `test03_rendering_backward` tests seem to be failing on the TeamCity build just for the `cuda_ad` variants of `PRBVolpath`. I think this is a miscompilation as the tests pass locally on my system and the same tests pass in `llvm_ad` variants. Any ideas?

dvicini:
Really nice work! Just as a heads up: I would assume that most Mitsuba-maintainers (including myself) will only have time to properly review this after the SIGGRAPH deadline (end of January).

Microno95:
> Really nice work! Just as a heads up: I would assume that most Mitsuba-maintainers (including myself) will only have time to properly review this after the SIGGRAPH deadline (end of January).

Thanks! And no worries, I expected as much with the SIGGRAPH deadline.

Microno95:
Hey @dvicini, any time opened up for reviewing this PR?

(As a side note, not sure why the tests are failing, they pass locally on a Windows setup with MSVC 2020 and Python 3.12 - I'll try and investigate this if I can get an Ubuntu setup running, difficult with WSL not supporting Optix.)

Microno95:
I've set up the repository in an Ubuntu 20.04 LTS docker container with Clang-18 and OptiX and python 3.13, and all the tests pass (except OptiX denoiser tests, likely because of flaky OptiX behaviour in WSL). I'm really not sure how to debug this issue. I will try LLVM-14 and GCC 9 on the same setup, hoping to replicate the issue, but it's strange that the results are so different. If it's a GPU/CUDA Architecture issue, that will be harder to debug as I only have access to RTX 30xx series GPUs.

I'm open to ideas at this point, happy to try anything.

johnson-c:
Just checking in where all this stands. Thanks everyone for making this available it is really a useful tool being adopted more and more in fusion research.

tloobyCFS:
Hey there.  First of all thanks for the amazing open source project.  We use this code a lot at Commonwealth Fusion Systems inside our digital twin workflows, and in plasma physics codes like [HEAT](https://github.com/plasmapotential/HEAT).  I am adding a ping here to give this thing some momentum.  Any updates or any chance this can be reviewed and merged?

Microno95:
Hi @tloobyCFS, @johnson-c,

Sorry for not replying sooner, from my understanding there is currently no one with the time in the EPFL-RGL group who is able to review this code for merging as it is a significant overhaul of various interfaces in the codebase.

I'm delighted to hear that CFS has also taken this codebase on-board for use in digital twins. It's really amazing to see! I want to ask that, for future reference as well in case anyone else chances upon this thread, you cite [Inverse rendering of fusion plasmas: inferring plasma composition from imaging systems](doi.org/10.1088/1741-4326/ad9ab5) when using the emissive rendering fork of the code (aside from the original Mitsuba 3 paper).

I will try and update the code to the latest version of Mitsuba when I have the time, and if you need the latest version urgently, drop me a message and I'd be happy to fast track that. Currently trying to balance postdoc research so this has been on the backburner.

Ekin

PR #680, opened 2023-04-18
Bitmap class: release GIL only after handling Python inputs

dvicini:
There is a small bug in the Bitmap Python bindings: the GIL should only be released after any Python operations are completed, not before. Otherwise it incurs undefined behavior due to Python internals updating without the GIL being held.


njroussel (approved):
Nice catch! Thank you :bowing_man:

PR #684, opened 2023-04-24
New shapes: cubic B-spline and linear curves

njroussel:
## Description

This PR adds two new shapes: `bsplinecurve` and `linearcurve`.

* Both shapes are meant to be used as aggregate objects holding many curves.
* Internally, they use OptiX and Embree for their intersection routines. Support for Mitsuba's built-in KD-tree has not been added. 
* Only the `bsplinecurve` is differentiable w.r.t. to its control points positions and radii.

In addition to the new plugins, this PR also has a few commits to refactor and simplify the OptiX shape handling.

Credits to @akaqyd and @ziyi-zhang for several parts of the implementation.

## Testing

Both new plugins have a new set of tests to check their intersection code and bounding boxes. In addition, the `bsplinecurve` has a set of tests to verify its differentiability. Unfortunately, due to bugs in the NVIDIA driver, theses tests are currently only set to run on CPU variants.


src/shapes/bsplinecurve.cpp
```diff
@@ -0,0 +1,816 @@
...
+        size_t segment_count = dr::width(m_indices);
+        UInt32 segment_id = dr::floor2int<UInt32>(v_global * segment_count);
+        Float v_local = v_global * segment_count - segment_id;
+
+        Point3f C;
+        Vector3f Cv, Cvv, Cvvv;
+        Float radius, rv, rvv;
+        std::tie(C, Cv, Cvv, Cvvv, radius, rv, rvv) =
+            cubic_interpolation(v_local, segment_id, active);
+
```
wjakob, line 712: Minor: it would be great to have more information on the source/further literature regarding this calculation.

src/python/python/ad/reparam.py
```diff
@@ -96,7 +96,7 @@ def _sample_warp_field(scene: mi.Scene,
 
     # Convert into a direction at 'ray.o'. When no surface was intersected,
     # copy the original direction
-    hit = si.is_valid()
+    hit = si.is_valid() & dr.any(dr.neq(si.p, ray.o))
```
Speierers, line 99: I am curious, why is this necessary? Is it specifically handling self-intersection with curves?

Btw, I wonder whether it wouldn't be necessary to use some epsilon.

src/render/shapegroup.cpp
```diff
@@ -40,8 +42,17 @@ MI_VARIANT ShapeGroup<Float, Spectrum>::ShapeGroup(const Properties &props) {
...
 #endif
-                m_has_meshes |= shape->is_mesh();
-                m_has_others |= !shape->is_mesh();
+                bool is_mesh = shape->is_mesh();
+                m_has_meshes |= is_mesh;
+
+                bool is_bspline = shape->is_mesh();
+                m_has_bspline_curves |= is_bspline;
+
+                bool is_linear = shape->is_mesh();
```
Speierers, line 51: ditto?

Speierers:
🚀

dvicini:
Nice work!

PR #687, opened 2023-04-25
Add bindings for Properties.as_string

Speierers:
This PR simply add Python bindings for the `Properties::as_string()` method.


njroussel (approved):
:rocket:

PR #688, opened 2023-04-25
Fix argument types in mi.luminance binding

Speierers:
This PR fixes the argument type in the bindings of `mi.luminance()`.


njroussel (approved):

njroussel:
Nice catch, thanks.

PR #690, opened 2023-04-25
Shape texture attribute

Speierers:
This PR adds the ability to set texture attributes to shape (not only `mesh`!), which can later be queried using the existing `eval_attribute()` method. This feature can be essential to design complex integrators that require to know shape attributes at render time. It can also become handy for quick prototyping, testing and visualization using Mitsuba.

This PR also add the `Shape::has_attribute()` method.

## Example

Here is a short example of how this new feature can be used:

```python
    mesh = mi.load_dict({
        "type" : "obj",
        "id" : "rect",
        "filename" : "resources/data/common/meshes/rectangle.obj",
        "attribute_1":  {
            "type": "bitmap",
            "filename" : "resources/data/common/textures/flower.bmp",
        }
    })

    si = mi.SurfaceInteraction3f()
    si.uv = mi.Point2f(0.5)

    color = mesh.eval_attribute('attribute_1', si),
```

### DIFF
--- a/include/mitsuba/python/docstr.h
+++ b/include/mitsuba/python/docstr.h
@@ -4294,6 +4294,8 @@ static const char *__doc_mitsuba_Mesh_embree_geometry = R"doc(Return the Embree 
 
 static const char *__doc_mitsuba_Mesh_ensure_pmf_built = R"doc()doc";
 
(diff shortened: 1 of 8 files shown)


njroussel (approved):
This looks good, thanks :smile:

(I've added a small commit to fix the failing test.)

PR #697, opened 2023-04-26
Use fresolver_append_path in curve tests

dvicini:
The curve tests were missing fresolver append path decorators to ensure that the test file is found when the test is run from a different path than the mitsuba root folder.


njroussel (approved):
Thank you, I keep forgetting them. I'm considering changing the CI so that it runs the tests from a non-root folder.

dvicini:
Yes that could make sense. Overall Mitsuba does a fairly thorough job at not hardcoding test paths, but indeed it's easy to forget.

PR #700, opened 2023-04-26
[✨feature] Add `mi.variant_context()`

leroyvn:
## Description

This PR adds a variant override context manager. It works as follows:

```python
mi.set_variant("scalar_rgb")

with mi.variant_context("llvm_rgb"):
    # Do some preprocessing faster with the LLVM backend
    ...

mi.render(my_scene)  # Will be rendered using "scalar_rgb"
```

I wrote this using a function-based context manager and checked that the initial variant is recovered if an exception occurs.

~~I also added a return value to `mi.set_variant()` so that the selected variant is known when multiple arguments are passed. This allows things like this (arguably of limited utility):~~

```python
mi.set_variant("scalar_rgb")

with mi.variant_context("llvm_rgb", "llvm_ad_rgb") as mi_variant:
    logging.debug("Using variant %s", mi_variant)
    ...

mi.render(my_scene)  # Will be rendered using "scalar_rgb"
```

## Testing

I added unit tests to the `test_util.py` suite.

## Checklist

### DIFF
--- a/src/python/python/tests/test_util.py
+++ b/src/python/python/tests/test_util.py
@@ -285,3 +285,30 @@ def test05_render_fwd_assert(variants_all_ad_rgb):
     with pytest.raises(Exception) as e:
         img = mi.render(scene)
         dr.forward_to(img)
+
+
+def test06_variant_context():
+    # Select the first variant which is not 'scalar_rgb'
+    for variant in mi.variants():
+        if variant != "scalar_rgb":
+            override_variant = variant
+            break
+    else:
+        pytest.skip("Only the 'scalar_rgb' variant was compiled.")
+
+    # Now, the test
+    mi.set_variant("scalar_rgb")
+
+    # The active variant is temporarily overridden
+    with mi.variant_context(override_variant):
+        assert mi.variant() == override_variant
+    assert mi.variant() == "scalar_rgb"
+
+    # The initial variant is restored if an exception is raised
+    try:
+        with mi.variant_context(override_variant):
+            assert mi.variant() == override_variant
+            raise RuntimeError
+    except RuntimeError:
+        pass
+    assert mi.variant() == "scalar_rgb"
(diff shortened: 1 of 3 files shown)


Speierers:
Thanks @leroyvn for this PR, this will be very useful to have!

Given that we often can `mi.set_variant()` at the end of a notebook cell, it can be a little distracting to always get it to output the selected variant. IMO I would rather not return it for this reason.

leroyvn:
I hadn't thought of that; I'll fix this today.

leroyvn:
I don't understand why the test fails for the macOS CI session. Thinking it was about the Python version (3.9 on the CI, 3.10 on my machine), I made an attempt in a fresh environment, but I cannot reproduce this. And I'm using the `contextmanager` decorator as recommended by the Python docs. Any idea of what could cause this?

src/python/python/tests/test_util.py
```diff
@@ -285,3 +285,21 @@ def test05_render_fwd_assert(variants_all_ad_rgb):
     with pytest.raises(Exception) as e:
         img = mi.render(scene)
         dr.forward_to(img)
+
+
+def test06_variant_context():
+    mi.set_variant("scalar_rgb")
+
```
njroussel, line 292:
```suggestion

if 'scalar_spectral' not in mi.variants():
    pytest.skip('Scalar spectral variant was not compiled.')
```

njroussel (changes requested):
This is great, thanks! (Sorry it took some time to take a look at it...)

The CI is failing because we don't compile the `scalar_spectral` variant on MacOS :slightly_frowning_face: . I've put a suggestion on how to ignore this for the test.
Or you can do something fancy where you lookup `mi.variants[1]`. 

After this small fix, it's ready to go :rocket:

src/python/python/tests/test_util.py
leroyvn: Thanks! I'll combine both suggestions and make a robust variant detection with an (unlikely) skip if only `scalar_rgb` is built.

leroyvn:
I also made the context manager re-raise any exception it encounters (I had accidentally committed a nasty silent pass).

njroussel (approved):
All good now, thanks!

PR #712, opened 2023-05-03
Detach vertex normals from previous AD graph on recomputation

### LINKED ISSUE #711: Vertex normal recomputation memory leak (?)
Hi,

I ran into an interesting issue today. It seems that the vertex normal update leaks memory when invoked in a differentiable computation. Here is a minimal reproducer, which prints the memory usage in each iteration. By setting `disable_normals = True` one can disable the normal recomputation, which removes the memory leak. 

```python
import mitsuba as mi
mi.set_variant('llvm_ad_rgb')
import drjit as dr

scene = mi.load_dict({'type': 'scene', 'mesh': {'type': 'cube'}})
params = mi.traverse(scene)

disable_normals = False
if disable_normals:
  params['mesh.vertex_normals'] = dr.full(mi.Float, 1.0, 0)

opt = mi.ad.Adam(1e-2)
key = 'mesh.vertex_positions'
opt[key] = params[key]
dr.enable_grad(opt[key])
for i in range(128):

  loss = dr.mean(dr.sqr(opt[key]))
  dr.backward(loss)
  opt.step()
  params.update(opt)

  dr.flush_malloc_cache()
  print('iter ', i)
  print('\n'.join(dr.whos_str().split('\n')[-13:]))

```

With a simple cube mesh the memory leak is somewhat negligible. However, it becomes more serious when using high-res geometry. Any idea what happens here? At first glance, the recompute_normals method in `mesh.cpp` doesn't look that unusual

dvicini:
Fixes #711 

The mesh class should make sure that the vertex normal field is detached from any prior AD graph on recomputation, otherwise this can cause an ever growing dependency chain during an optimization.

Detaching the vertex normals here doesn't change anything about the result of the computation, as either way in this function the vertex normals are being overwritten. But since the write happens through scatter operations, DrJit cannot know that the previously connected AD graph isn't used anymore.

### DIFF
--- a/src/render/mesh.cpp
+++ b/src/render/mesh.cpp
@@ -339,6 +339,11 @@ MI_VARIANT void Mesh<Float, Spectrum>::recompute_vertex_normals() {
 
         normals = dr::normalize(normals);
 
+        // Disconnect the vertex normal buffer from any pre-existing AD
+        // graph. Otherwise an AD graph might be unnecessarily retained
+        // here, despite the following lines re-initializing the normals.
+        dr::disable_grad(m_vertex_normals);
+
         UInt32 ni = dr::arange<UInt32>(m_vertex_count) * 3;
         for (size_t i = 0; i < 3; ++i)
             dr::scatter(m_vertex_normals,


Speierers:
Nice catch! Indeed the scatter won't detach the vertex normal from its previous AD graph.

One minor comment otherwise this looks good to me!

src/render/mesh.cpp
```diff
@@ -338,7 +338,7 @@ MI_VARIANT void Mesh<Float, Spectrum>::recompute_vertex_normals() {
         // --------------------- Kernel 2 starts here ---------------------
 
         normals = dr::normalize(normals);
-
+        m_vertex_normals = dr::detach(m_vertex_normals);
```
Speierers, line 341:
```suggestion
       dr::disable_grad(m_vertex_normals);
```

dvicini: You are right, using disable_grad is better here. Also added a comment explaining why its needed

njroussel (approved):
Ohhh indeed!

Thanks :smile: :+1:

PR #721, opened 2023-05-08
Add support for per-emitter sampling weight

dvicini:
Hi,

I've recently had the need for more customized light sampling (instead of uniform sampling of all available emitters). Out of the box, Mitsuba 3 only supports uniform emitter sampling, which is somewhat limiting, in particular for applications that require on-the-fly modification of light sources. Such modifications are essential for inverse rendering problems involving some form of active illumination (e.g., see also https://github.com/mitsuba-renderer/mitsuba3/discussions/402). Currently the only way to do this seems to be to write a custom integrator. This is always an option, but a relatively heavy solution to just change the sampling of emitters (It effectively requires a duplicating every integrator that one wants to use). I don't think there is another way currently, or at least I couldn't figure out a good way to do it (e.g., a custom emitter plugin that internally handles these sampling decisions doesn't quite work, as we can't attach one emitter to multiple shapes and can't replace a shape's emitter easily after scene loading).

This PR adds a per-emitter `sampling_weight`. The design for this follows Mitsuba 0.6, which also had this weight. The code falls back to the fast uniform sampling if all sampling weights are equal to 1, so that there shouldn't be a performance regression for the default case. I also added some rudimentary tests to `test_scene.py`.

In the future, it could be interesting to factor out all the emitter sampling into a new plugin type (e.g., `EmitterSampler`), to allow users to more selectively replace functionality that's currently hardcoded in `scene.cpp`. But for now I think just re-introducing the sampling weight is probably the easiest solution, which doesn't modify any major interfaces and offers a clear path for users without a deep knowledge of the inner workings of the system. Eventually, introducing a special plugin that handles these sampling decisions could make sense, as it would allow customization at the user level without the need to add much code to Mitsuba itself.

(still slightly work in progress)

### DIFF
--- a/include/mitsuba/render/emitter.h
+++ b/include/mitsuba/render/emitter.h
@@ -69,6 +69,14 @@ public:
 
     void traverse(TraversalCallback *callback) override;
 
+    void parameters_changed(const std::vector<std::string> &keys = {}) override;
+
+    /// Return whether the emitter parameters have changed
+    bool dirty() const { return m_dirty; }
+
+    /// Mark that the emitters's parameters have changed
+    void mark_dirty() { m_dirty = true; }
+
     DRJIT_VCALL_REGISTER(Float, mitsuba::Emitter)
 
     MI_DECLARE_CLASS()
@@ -83,6 +91,9 @@ protected:
 
     /// Sampling weight
     ScalarFloat m_sampling_weight;
+
+    /// True if the emitters's parameters have changed
+    bool m_dirty;
 };
 
 MI_EXTERN_CLASS(Emitter)
--- a/include/mitsuba/render/emitter.h
+++ b/include/mitsuba/render/emitter.h
@@ -53,6 +53,7 @@ template <typename Float, typename Spectrum>
 class MI_EXPORT_LIB Emitter : public Endpoint<Float, Spectrum> {
 public:
     MI_IMPORT_BASE(Endpoint, m_shape)
+    MI_IMPORT_TYPES()
 
     /// Is this an environment map light emitter?
     bool is_environment() const {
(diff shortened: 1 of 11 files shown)


njroussel:
Hi @dvicini 

This is looking good! I don't have any comments right now. Please ping me again once its' fully ready. Or are you finished now?

------

>  Eventually, introducing a special plugin that handles these sampling decisions could make sense, as it would allow customization at the user level without the need to add much code to Mitsuba itself.

We had discussed this at some point and it seemed like a great idea. However, this would come at the cost of an extra virtual function call in scalar mode which we'd like to avoid. I think this was the sole reason why we dropped this idea. I think this could still be something we offer, at least, in JIT variants. For scalar variants, it would at least be nice to know the cost of the extra vcall.

dvicini:
Thanks for taking a look. I will take another pass to make sure it's all correct and then let you know before merging :)

Agree with the undesirable cost for the extra VCall, this is the main downside of that approach. For the JIT variants it's all fine because either way it can be inlined. Maybe there could be some static workaround for the default case when no specialized sampling is used. Either way, I think for now the sampling weight is the best solution due to not creating any significant overhead.

dvicini:
Updated with a mark_dirty flag to tag updated emitters and cleaned up the PR overall. This is ready for review / merge now

include/mitsuba/render/emitter.h
```diff
@@ -74,6 +88,12 @@ class MI_EXPORT_LIB Emitter : public Endpoint<Float, Spectrum> {
 protected:
     /// Combined flags for all properties of this emitter.
     uint32_t m_flags;
+
+    /// Sampling weight
+    ScalarFloat m_sampling_weight;
+
+    /// True if the emitters's parameters have changed
+    bool m_dirty;
```
njroussel, line 96: Is this ever initialized? Or reset to false after the update?

src/render/scene.cpp
```diff
@@ -334,6 +370,18 @@ MI_VARIANT void Scene<Float, Spectrum>::parameters_changed(const std::vector<std
...
+    // the emitter sampling distribution.
+    bool emitters_changed = false;
+    for (auto &e : m_emitters) {
+        if (e->dirty()) {
+            emitters_changed = true;
+            break;
+        }
+    }
+    if (emitters_changed)
+        update_emitter_sampling_distribution();
```
njroussel, line 384: "Unmark" the emitters as dirty here.

njroussel (changes requested):
A couple of comments regarding the `mark_dirty` implementation.

Also, I can't put this comment on the correct line: 
In `point.cpp` the `parameters_changed()` method should also call the `Base::parameters_changed()`.

dvicini:
Good catch! I fixed both point.cpp and the mark_dirty stuff

PR #726, opened 2023-05-11
Improve numerical precision in prb.py 

Speierers:
In the last phase of PRB, local contribution is substracted to the the total path contribution at every bounce. This commit ensures the substracted values are exactly equivalent to the ones computed in the previous phases using the replace_grad routine when re-attaching the detached emitter sample.


njroussel (approved):
Yes! Thanks you :+1:

PR #728, opened 2023-05-11
Add eval_attribute methods to BSDF

Speierers:
# Overview

Similar to #690, this PR adds to BSDFs the ability to expose some of their internal attributes. The interface is identical to `Shape`. This can be useful for prototyping, and also when other parts of the renderer need to access such attributes (e.g. within a custom integrator).

# Testing

If you are interested in merging this PR onto the official codebase, I can add a few unit tests to test this new feature.

### DIFF
--- a/src/render/bsdf.cpp
+++ b/src/render/bsdf.cpp
@@ -40,6 +40,42 @@ MI_VARIANT Spectrum BSDF<Float, Spectrum>::eval_diffuse_reflectance(
     return eval(ctx, si, wo, active) * dr::Pi<Float>;
 }
 
+
+MI_VARIANT typename BSDF<Float, Spectrum>::Mask
+BSDF<Float, Spectrum>::has_attribute(const std::string& /*name*/, Mask /*active*/) const {
+    return false;
+}
+
+MI_VARIANT typename BSDF<Float, Spectrum>::UnpolarizedSpectrum
+BSDF<Float, Spectrum>::eval_attribute(const std::string & name,
+                                      const SurfaceInteraction3f & /*si*/,
+                                      Mask /*active*/) const {
+    if constexpr (dr::is_jit_v<Float>)
+        return 0.f;
+    else
+        Throw("Invalid attribute requested %s.", name.c_str());
+}
+
+MI_VARIANT Float
+BSDF<Float, Spectrum>::eval_attribute_1(const std::string& name,
+                                        const SurfaceInteraction3f & /*si*/,
+                                        Mask /*active*/) const {
+    if constexpr (dr::is_jit_v<Float>)
+        return 0.f;
+    else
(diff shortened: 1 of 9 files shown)


dvicini:
Nice idea, I like the added flexibility! +1 from me on merging this with some tests

Speierers:
In my latest commit, the BSDF now uses the `traverse()` method to automatically find the attributes to evaluate. With this the added methods (e.g. `eval_attribute`) won't need to be overridden for the existing BSDFs.

src/render/bsdf.cpp
```diff
@@ -40,6 +41,138 @@ MI_VARIANT Spectrum BSDF<Float, Spectrum>::eval_diffuse_reflectance(
...
+                                        const SurfaceInteraction3f & si,
+                                        Mask active) const {
+    bool found = false;
+    Color3f result;
+
+    auto put_object = [&](const std::string &name_, Object *obj) {
+        if (name == name_) {
+            Texture *texture = dynamic_cast<Texture *>(obj);
+            if (texture) {
+                result = texture->eval_1(si, active);
```
dvicini, line 152: eval_3?

Speierers: Indeed!

dvicini:
The C++ code looks good to me now. The only missing things would be to add some tests, both calling eval_attribute on existing BSDF and maybe a test for a Python-defined BSDF that exposes some custom attributes.

dvicini:
Really cool, looks good to me 👍

dvicini:
Shall we merge this @njroussel @Speierers ?

Speierers:
@njroussel do you think we can merge this or are those changes too intrusive for the official codebase?

njroussel (commented):
Sorry -- at some point I must have misclicked and manually unwatched this PR, I wasn't getting any notifications.


This looks great. I think it will be nice to extend the `aov` integrator to automatically support these attributes too.

One suggestion:
Instead of having a default value of 0 for the result whenever the attribute is not present, should we rather just return a pair that includes a boolean mask?  The actual value zero might be ambiguous for some attribute.

Speierers:
> Instead of having a default value of 0 for the result whenever the attribute is not present, should we rather just return a pair that includes a boolean mask? The actual value zero might be ambiguous for some attribute.

@njroussel I agree simply returning zero in that case isn't the best solution. But the idea here was to match the existing attribute API of the `Shape` plugins. Also, that's the reason why I added the `has_attribute` method, which the user can use to actually check whether the attribute really exists or not.

> This looks great. I think it will be nice to extend the aov integrator to automatically support these attributes too.

Indeed that would be very cool! But I would do this in a different PR.

njroussel:
Ohhh, I hadn't thought about the `has_attribute`. All good then!

PR #731, opened 2023-05-12
Vectorize BSDFContext

Speierers:
## Description

This PR adds vectorization to the `BSDFContext` datastructure passed to the BSDF routines (e.g. `eval()`, `sample()`, `pdf()`). This allows the callee to specify the BSDF components and event types to be sampled/evaluated on a per-lane basis. This is very useful for implementing non-conventional integrators for the vectorized variants of Mitsuba (e.g. light path expression, ...).

This PR adapts the code of all BSDFs to use masking logic to handle the non-scalar conditions based on the BSDF context object.

## Testing

On top of the current test suite which already tests the activation/de-activation of the BSDF components using the BSDFConext, I added a simple test in `test_dielectric.py` to test it in a vectorized way.


wjakob:
Just a quick note (also for Nicolas): I am a little skeptical about this change and the cost/complexity that it adds. Let's discuss further after the SIGGRAPH Asia deadline.

Speierers:
Pinging @wjakob and @njroussel for discussing this PR further.

njroussel:
Thanks for the reminder :smile: 

I don't have any strong opinions on this. I personally don't feel like this adds a whole lot of additional complexity to the renderer. 

The only reason why I wouldn't want to merge this PR is because we don't have an actual supported use case for it (yet - maybe LPEs). However, there have been discussions/threads of people needing more "specialization" per ray and this would be a step towards that. 

Happy to hear what your initial thoughts were @wjakob.

Speierers:
After discussing with @njroussel , it was decided not to merge this PR into `master` due to the invasive nature of the changes and limited clarity on the use case for the vectorized `BSDFContext` datastructure. We'll keep the PR open for reference and potential future discussions on merging it into `master` if the need arises.

PR #738, opened 2023-05-17
Large steps support & shape optimization tutorial

bathal1:
## Description

This PR brings the following changes:
- Added support for the algorithm from the "[Large Steps in Inverse Rendering of Geometry](http://rgl.epfl.ch/publications/Nicolet2021Large)" paper in Mitsuba. It currently only support meshes.
- Added some additional to `Mesh::parameters_changed` to avoid unwanted behavior when updating a given mesh's tesselation, e.g. through remeshing
- Added a "Shape optimization" tutorial, that ewplains how to use the "Large Steps" method for shape optimization.

This would conclude the discussion in #600.

### DIFF
--- a/src/python/python/ad/largesteps.py
+++ b/src/python/python/ad/largesteps.py
@@ -1,0 +1,167 @@
+from __future__ import annotations as __annotations__ # Delayed parsing of type annotations
+
+import mitsuba as mi
+import drjit as dr
+import numpy as np
+
+def mesh_laplacian(n_verts, faces, lambda_):
+    """
+    Compute the index and data arrays of the (combinatorial) Laplacian matrix of
+    a given mesh.
+    """
+    # Neighbor indices
+    ii = faces[:, [1, 2, 0]].flatten()
+    jj = faces[:, [2, 0, 1]].flatten()
+    adj = np.unique(np.stack([np.concatenate([ii, jj]), np.concatenate([jj, ii])], axis=0), axis=1)
+    adj_values = np.ones(adj.shape[1], dtype=np.float64) * lambda_
+
+    # Diagonal indices, duplicated as many times as the connectivity of each index
+    diag_idx = np.stack((adj[0], adj[0]), axis=0)
+
+    diag = np.stack((np.arange(n_verts), np.arange(n_verts)), axis=0)
+
+    # Build the sparse matrix
+    idx = np.concatenate((adj, diag_idx, diag), axis=1)
+    values = np.concatenate((-adj_values, adj_values, np.ones(n_verts)))
+
+    return idx, values
+
+class SolveCholesky(dr.CustomOp):
+    """
+    DrJIT custom operator to solve a linear system using a Cholesky factorization.
+    """
+
+    def eval(self, solver, u):
+        self.solver = solver
+        x = dr.empty(dr.cuda.TensorXf, shape=u.shape)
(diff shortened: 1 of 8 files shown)


src/python/python/ad/largesteps.py
```diff
@@ -0,0 +1,175 @@
...
+    return idx, values
+
+class SolveCholesky(dr.CustomOp):
+    """
+    DrJIT custom operator to solve a linear system using a Cholesky factorization.
+    """
+
+    def eval(self, solver, u):
+        self.solver = solver
+        x = dr.empty(dr.cuda.TensorXf, shape=u.shape)
```
DoeringChristian, line 36: Changing the type to `mi.TensorXf` allows this to work in llvm modes as well i.e.
```python
        x = dr.empty(mi.TensorXf, shape=u.shape)
```

src/python/python/ad/largesteps.py
```diff
@@ -0,0 +1,175 @@
...
+        solver.solve(u, x)
+        return mi.TensorXf(x)
+
+    def forward(self):
+        x = dr.empty(mi.TensorXf, shape=self.grad_in('u').shape)
+        self.solver.solve(self.grad_in('u'), x)
+        self.set_grad_out(x)
+
+    def backward(self):
+        x = dr.empty(dr.cuda.TensorXf, shape=self.grad_out().shape)
```
DoeringChristian, line 46: Changing the type to `mi.TensorXf` allows this to work in llvm modes as well i.e.
```python
        x = dr.empty(mi.TensorXf, shape=self.grad_out().shape)

```

DoeringChristian:
Hi @bathal1,

Thanks for your implementation it is definitively the way to do it properly :sweat_smile:.
I have tested your implementation as there where some problems with my implementation using the `pytorch` interface (see [here](https://github.com/mitsuba-renderer/mitsuba3/discussions/600#discussioncomment-6035224)).
And I have two notes:
First, as I also tested it in `llvm` modes I found two lines where you used `dr.cuda.TensorXf` which limits the use to `cuda` variants. In my case I changed these to use the Mitsuba type i.e. `mi.TensorXf`. I added comments [here](https://github.com/mitsuba-renderer/mitsuba3/pull/738/files#r1212294268) and [here](https://github.com/mitsuba-renderer/mitsuba3/pull/738/files#r1212294635).
Second, the problem described in the discussion #600 also seems to affect your implementation (only for `cuda` variants and with the `prb` integrator).
The problem was introduced with [this](https://github.com/mitsuba-renderer/drjit-core/compare/f657fe0c8c10f5d0aa60156905ffa3f760684b52...21bdbebfde8a7122f4d82e0a86df155bfb43ffb6#) change or the problem exists in that integrator and only became noticeable with the change. It seems to have something to do with `opaque` variables (they might need some documentation) and it might affect other projects that use the `prb` integrator.

njroussel:
Hi @DoeringChristian 

Thanks for your review :+1:, I've been editing and improving the notebook/tutorial, and haven't made my way to the `largesteps.py` file yet. I should be done later today.

Regarding the flipped normals issue raised in #600, I've answered over there.

njroussel (approved):
Thank you everyone!

I'm happy with the current state of this PR/feature. We can revisit this in the future if we need to add support for textures or any other arbitrary structure actually.

PR #748, opened 2023-05-23
Add support for spectral phase functions

Speierers:
This PR modifies the interface of the `PhaseFunction` class to support spectral evaluation. This is done by replacing the `eval()` method by `eval_pdf()` which returns both a spectrum value and the corresponding sampling PDF. `PhaseFunction::sample()` now also returns a sampling weight similar to the one returned by `BSDF::sample()`.

The `volpath`, `volpathmis` and `prbvolpath` integrators were modified to account for theses changes.

On top of this, this PR also renames `Medium::tr_and_pdf` into `Medium::transmission_eval_pdf` to be more consistent with other plugin interfaces.

### DIFF
--- a/include/mitsuba/render/phase.h
+++ b/include/mitsuba/render/phase.h
@@ -22,19 +22,7 @@ enum class PhaseFunctionFlags : uint32_t {
     Microflake  = 0x04
 };
 
-constexpr uint32_t operator|(PhaseFunctionFlags f1, PhaseFunctionFlags f2) {
-    return (uint32_t) f1 | (uint32_t) f2;
-}
-constexpr uint32_t operator|(uint32_t f1, PhaseFunctionFlags f2) { return f1 | (uint32_t) f2; }
-constexpr uint32_t operator&(PhaseFunctionFlags f1, PhaseFunctionFlags f2) {
-    return (uint32_t) f1 & (uint32_t) f2;
-}
-constexpr uint32_t operator&(uint32_t f1, PhaseFunctionFlags f2) { return f1 & (uint32_t) f2; }
-constexpr uint32_t operator~(PhaseFunctionFlags f1) { return ~(uint32_t) f1; }
-constexpr uint32_t operator+(PhaseFunctionFlags e) { return (uint32_t) e; }
-template <typename UInt32> constexpr auto has_flag(UInt32 flags, PhaseFunctionFlags f) {
-    return (flags & (uint32_t) f) != 0;
-}
+MI_DECLARE_ENUM_OPERATORS(PhaseFunctionFlags)
 
 /**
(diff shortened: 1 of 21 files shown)


dvicini:
Hi Sebastien,
At a high level, this looks good to me. Spectral phase functions was a somewhat long-standing feature request I think, so adding this is good.

Just two small questions/comments:
1) I would call the medium function `transmittance_eval_pdf` (instead of `transmission`)
2) Are we currently actually testing all the variants of the volpath? (simple, MIS, PRB) Just to make sure we don't accidentally introduce some regressions there.

Speierers:
Thanks for your feedbacks @dvicini, I renamed the method to `transmittance_eval_pdf`.

And yes, we do test all variants of the volumetric integrators in `test_renders.py`.

include/mitsuba/render/phase.h
```diff
@@ -141,16 +129,16 @@ class MI_EXPORT_LIB PhaseFunction : public Object {
      *     A uniformly distributed sample on \f$[0,1]^2\f$. It is
      *     used to generate the sampled direction.
      *
-     * \return A sampled direction wo
+     * \return A sampled direction wo and the corresponding weight and PDF
```
njroussel, line 132:
```suggestion
     * \return A sampled direction wo and its corresponding weight and PDF
```

include/mitsuba/render/phase.h
```diff
@@ -141,16 +129,16 @@ class MI_EXPORT_LIB PhaseFunction : public Object {
...
-                                              const MediumInteraction3f &mi,
-                                              Float sample1, const Point2f &sample2,
-                                              Mask active = true) const = 0;
+    virtual std::tuple<Vector3f, Spectrum, Float> sample(const PhaseFunctionContext &ctx,
+                                                         const MediumInteraction3f &mi,
+                                                         Float sample1, const Point2f &sample2,
+                                                         Mask active = true) const = 0;
     /**
-     * \brief Evaluates the phase function model
+     * \brief Evaluates the phase function model value and pdf
```
njroussel, line 139:
```suggestion
     * \brief Evaluates the phase function model value and PDF
```

include/mitsuba/render/phase.h
```diff
@@ -165,10 +153,12 @@ class MI_EXPORT_LIB PhaseFunction : public Object {
      * \param wo
      *     An outgoing direction to evaluate.
      *
-     * \return The value of the phase function in direction wo
+     * \return The value and the sampling pdf of the phase function in direction wo
```
njroussel, line 156:
```suggestion
     * \return The value and the sampling PDF of the phase function in direction wo
```

njroussel:
Thank you!

(I caught up with Miguel, nothing really relevant for this PR, just some docs that I will add later on)

PR #750, opened 2023-05-24
Expose m_film member in Sensor python bindings

Speierers:
Short patch that adds bindings to the underlying Film object in sensors. For instance this can be used to access the film object from a Python custom sensor plugin.

### DIFF
--- a/src/render/python/sensor_v.cpp
+++ b/src/render/python/sensor_v.cpp
@@ -78,6 +78,7 @@ public:
 
     using Sensor::m_needs_sample_2;
     using Sensor::m_needs_sample_3;
+    using Sensor::m_film;
 };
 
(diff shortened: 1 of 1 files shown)


njroussel:
Hi @Speierers 

How is this different from using [`Sensor::film()`](https://mitsuba.readthedocs.io/en/stable/src/api_reference.html#mitsuba.Sensor.film) which already has Python bindings?

Speierers:
The `m_film` bindings are read and **write**, while `film()` can't be used to set the film instance to be used during rendering.

njroussel:
Oh, I see. I hadn't considered that you might want to change the film type and not just its parameters. 
LGTM :+1:

PR #751, opened 2023-05-24
Add SensorPtr::sample_ray_differential bindings

Speierers:
Short patch to add missing bindings for `SensorPtr::sample_ray_differential`.


njroussel (approved):
Thank you :rocket:

PR #752, opened 2023-05-24
Fix MSVC warnings

Speierers:
This patch fixes various MSVC warnings introduced recently in the codebase.


njroussel (approved):
Great, thanks !

PR #768, opened 2023-06-08
Blender exporter refactor

bathal1:
## Description

This PR addresses compatibility issues caused by data layout changes in Blender's mesh representation for the exporter add-on. With these changes, the exporter should now support versions 3.4 and 3.5 of Blender, which should make it easier to support the upcoming LTS version.

The mesh layout in blender is progressively moving from an array of structs to a struct of arrays format (more detail [here](https://projects.blender.org/blender/blender/issues/95965), so we need to reflect those changes when interpreting blender data pointers in the mesh exporter.

This PR fixes [this](https://github.com/mitsuba-renderer/mitsuba-blender/issues/58) issue. In order for these changes to be fully reflected in the exporter, we would need to have a new release of the Mitsuba PyPI package and update the dependency in the addon.

### DIFF
--- a/src/shapes/blender.cpp
+++ b/src/shapes/blender.cpp
@@ -142,7 +142,9 @@ public:
         const bool *sharp_faces = reinterpret_cast<const bool *>(props.get<int64_t>("sharp_face", 0));
 
         // The type of vertex buffer will depend on the version of blender used.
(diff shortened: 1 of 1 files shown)


njroussel (approved):
Looks good to me! :+1:

Issue #771, opened 2023-06-09
Nested area emitter segfaults


dvicini:
Hi,

I ran into an edge case using a nested area emitter defined in Python. I can render a single scene using this emitter just as expected. However, if I subsequently instantiate and render another scene in the same process, the VCall tracing fails. I get a segmentation fault when the nested area emitter wants to call into the associated shape (inside the `sample_direction` function). 

Looking into this some more, it seems that the VCall tracing both times uses the same `m_shape` pointer. The first time we render a scene this is the correct pointer to the associated shape. But the second time the `m_shape` pointer points to a now invalid address. 

So somehow the VCall tracing seems to use an outdated version of the function, which bakes in the old pointer.

Any idea why this would happen?

Update: On second thought - could it be that the recording with the outdated shape pointer is happening due to some dangling reference to the original area emitter? I.e., the original area emitter still is somewhere in the system, while the shape was already cleaned up? (the area emitter holds a pointer and not a ref<Shape>). In that case we should figure out where that reference comes from...

Here a minimal reproducer. The first call to `f` succeeds and the second one produces a segfault.
```python
import mitsuba as mi
import drjit as dr

mi.set_variant('llvm_ad_rgb')

class MyEmitter(mi.Emitter):

  def __init__(self, props):
    mi.Emitter.__init__(self, props)
    self.nested_emitter = props['nested_emitter']
    self.m_flags = self.nested_emitter.flags()

  def prepare(self):
    self.nested_emitter.set_shape(self.shape())

  def eval(self, si, active):
    return self.nested_emitter.eval(si, active)

  def sample_direction(self, it, sample, active):
    return self.nested_emitter.sample_direction(it, sample, active)
    
  def eval_direction(self, it, ds, active):
    return self.nested_emitter.eval_direction(it, ds, active)

  def pdf_direction(self, it, ds, active):
    return self.nested_emitter.pdf_direction(it, ds, active)

  def bbox(self):
    return self.nested_emitter.bbox()

  def to_string(self):
    return f"""MyEmitter[
   nested_emitter = {self.nested_emitter}
                         ]"""


mi.register_emitter('my_emitter', lambda props: MyEmitter(props))

def f():
    scene = mi.load_dict(
        {
            'type': 'scene',
            'integrator': {'type': 'direct'},
            'light': {
                'type': 'sphere',
                'emitter': {
                    # 'type': 'area',
                    'type': 'my_emitter',
[...]

PR #776, opened 2023-06-14
Fix Python plugin reference leak

dvicini:
There is a known issue in Mitsuba that a reference to the Python object is leaked when a Python-defined plugin is instantiated.

The original reason we had to leak this reference was to keep the Python object alive beyond the scope of the initialization call in `python.h`. The problem is that due to this leak, the Python object is never deallocated and may end up in a partially invalid state (e.g., see the nested emitter example in #771).

Ideally, we could fully transfer the ownership of the Python object to C++ and Mitsuba's `ref` object. This doesn't seem to be possible in pybind11 and as far as I can tell is a known limitation. It will be fixed in nanobind (see the [docs](https://nanobind.readthedocs.io/en/latest/ownership.html#intrusive-reference-counting)).

This PR implements a relatively straightforward fix that should work until the transition to Nanobind is complete. The solution that seems to work is to keep a Python object handle in a cleanup callback lambda, and then explicitly invoke the cleanup call when the object's reference count decreases from 2 to 1 (i.e., the only remaining reference is the Python object). This isn't very elegant but appears to solve the problem on my end.

Not sure if we want to merge this, but I thought I would share this since the transition to Nanobind might still take some time (?). The leaked reference ended up causing some unforeseen issues on my end, so maybe others can benefit from a fix as well.

### DIFF
--- a/include/mitsuba/core/object.h
+++ b/include/mitsuba/core/object.h
@@ -118,6 +118,11 @@ public:
      */
     virtual std::string to_string() const;
 
+    /// Set a cleanup callback for Python-defined plugins.
+    void set_py_cleanup(const std::function<void()> &callback) {
+        m_py_cleanup_callback = callback;
+    }
+
 protected:
     /** \brief Virtual protected deconstructor.
      * (Will only be called by \ref ref)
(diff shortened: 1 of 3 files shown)


njroussel:
For anyone that comes across this PR, we don't plan on merging these changes. However, we encourage anyone to incorporate this change into their own fork if they are running into issues with leaked references of custom plugins.

As mentioned in the PR description, an upcoming release should properly fix these issues. Until then, I will keep this PR open for visibility.

PR #778, opened 2023-06-16
SDFGrid: Adding Embree support

rtabbara:
This PR adds support for using SDFGrid shapes when targeting llvm/scalar variants. Here, accelerated grid traversal is reliant on Embree

The implementation of voxel intersection handling is based off of the existing CUDA side work, but has been re-expressed to support packet types. Of particular note:

* Embree relies on callbacks to supply custom geometry bboxes per primitive (voxel). This is in contrast to OptiX, where we can immediately supply a single buffer of all bounding boxes. As such, similar to mesh handling, we need to modify the interface of `Shape::ray_intersect_preliminary/Shape::ray_test` to optionally provide a primitive index to determine which voxel the ray is intersecting.
* Similar to the OptiX implementation, we filter out any empty voxels and maintain a mapping from primitive index to the voxel position.
* The sdfgrid tests scope have been extended to test all variants

Outstanding things left:

- [X] Remove references to cubic interpolation in `SDFGrid` 
- [X] Remove Falcao method for generating normals.
- [X] Fill out documentation for SDFGrid including image examples
- [X] Fix instancing support on CUDA (llvm appears to work)
- [X] Add support for sdf grid data source similar to volume grid

### DIFF
--- a/include/mitsuba/render/shape.h
+++ b/include/mitsuba/render/shape.h
@@ -137,9 +137,8 @@ public:
      *     Index of the primitive to be intersected
      */
     virtual PreliminaryIntersection3f ray_intersect_preliminary(const Ray3f &ray,
-                                                                Mask active = true,
-                                                                ScalarIndex prim_index = 0) const;
-
+                                                                ScalarIndex prim_index = 0,
+                                                                Mask active = true) const;
     /**
      * \brief Fast ray shadow test
      *
@@ -156,7 +155,7 @@ public:
      * \param prim_index
      *     Index of the primitive to be intersected
      */
-    virtual Mask ray_test(const Ray3f &ray, Mask active = true, ScalarIndex prim_index = 0) const;
+    virtual Mask ray_test(const Ray3f &ray, ScalarIndex prim_index = 0, Mask active = true) const;
 
     /**
      * \brief Compute and return detailed information related to a surface interaction
@@ -237,11 +236,11 @@ public:
         using Ray3fP##N   = Ray<Point3fP##N, Spectrum>;                         \
         virtual std::tuple<FloatP##N, Point2fP##N, UInt32P##N, UInt32P##N>      \
         ray_intersect_preliminary_packet(const Ray3fP##N &ray,                  \
-                                         MaskP##N active = true,                \
-                                         ScalarIndex prim_index = 0) const;     \
+                                         ScalarIndex prim_index = 0,            \
+                                         MaskP##N active = true) const;         \
         virtual MaskP##N ray_test_packet(const Ray3fP##N &ray,                  \
-                                         MaskP##N active = true,                \
-                                         ScalarIndex prim_index = 0) const;
+                                         ScalarIndex prim_index = 0,            \
+                                         MaskP##N active = true) const;
 
     MI_DECLARE_RAY_INTERSECT_PACKET(4)
     MI_DECLARE_RAY_INTERSECT_PACKET(8)
@@ -598,18 +597,18 @@ NAMESPACE_END(mitsuba)
     using typename Base::Ray3fP##N;                                                         \
     std::tuple<FloatP##N, Point2fP##N, UInt32P##N, UInt32P##N>                              \
     ray_intersect_preliminary_packet(                                                       \
-        const Ray3fP##N &ray, MaskP##N active, ScalarIndex prim_index) const override {     \
+        const Ray3fP##N &ray, ScalarIndex prim_index, MaskP##N active) const override {     \
         (void) ray; (void) active;                                                          \
         if constexpr (!dr::is_cuda_v<Float>)                                                \
-            return ray_intersect_preliminary_impl<FloatP##N>(ray, active, prim_index);      \
+            return ray_intersect_preliminary_impl<FloatP##N>(ray, prim_index, active);      \
         else                                                                                \
             Throw("ray_intersect_preliminary_packet() CUDA not supported");                 \
     }                                                                                       \
-    MaskP##N ray_test_packet(const Ray3fP##N &ray, MaskP##N active, ScalarIndex prim_index) \
+    MaskP##N ray_test_packet(const Ray3fP##N &ray, ScalarIndex prim_index, MaskP##N active) \
         const override {                                                                    \
         (void) ray; (void) active;                                                          \
         if constexpr (!dr::is_cuda_v<Float>)                                                \
-            return ray_test_impl<FloatP##N>(ray, active, prim_index);                       \
+            return ray_test_impl<FloatP##N>(ray, prim_index, active);                       \
         else                                                                                \
             Throw("ray_intersect_preliminary_packet() CUDA not supported");                 \
     }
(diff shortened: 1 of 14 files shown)


src/shapes/sdfgrid.cpp
```diff
@@ -494,57 +601,349 @@ class SDFGrid final : public SDF<Float, Spectrum> {
...
+
+        FloatP t_near = t_beg;
+        FloatP t_far  = t_end;
+
+        FloatP f_root_0 = eval_sdf_t(root_0);
+        FloatP f_root_1 = eval_sdf_t(root_1);
+
+        MaskP root_0_valid = t_near <= root_0 && root_0 <= t_far;
+
+        dr::masked(t_far, has_derivative_roots && root_0_valid &&
```
wjakob, line 815: `dr::masked` is only needed if the left hand side is potentially *not* a dr.jit array.
You may prefer the slice assignment syntax for Dr.Jit arrays:

`array[mask] = ...;`.

wjakob (approved):

src/shapes/sdfgrid.cpp
```diff
@@ -562,70 +961,95 @@ class SDFGrid final : public SDF<Float, Spectrum> {
...
+            aabbs_ptr         = aabbs;
+            voxel_indices_ptr = voxel_indices;
+        }
+
+        return { aabbs_ptr, voxel_indices_ptr, count };
     }
 
     /// Computes the gradient for a specific gradient
-    Vector3f voxel_grad(const Point3f& p, const Point3i& voxel_index) const {
+    Vector3f voxel_grad(const Point3f &p, const Point3i &voxel_index) const {
```
ziyi-zhang, line 1029: (unrelated to the embree support)
Would this be more efficient?
```
Vector3f voxel_grad(const Point3f& p, const Point3i& voxel_index, Mask active = true) const {
        Float f000, f001, f010, f011, f100, f101, f110, f111;
        dr::Array<Float *, 8> fetch_values;
        fetch_values[0] = &f000;
        fetch_values[1] = &f100;
        fetch_values[2] = &f010;
        fetch_values[3] = &f110;
        fetch_values[4] = &f001;
        fetch_values[5] = &f101;
        fetch_values[6] = &f011;
        fetch_values[7] = &f111;
        m_grid_texture.eval_fetch(rescale_point(p), fetch_values, active);

        Point3f p_local = p * m_resolution - Point3f(voxel_index);
        Float y0 = dr::lerp(f100 - f000, f110 - f010, p_local.y());
        Float y1 = dr::lerp(f101 - f001, f111 - f011, p_local.y());
        Float dx = dr::lerp(y0, y1, p_local.z());
        Float x0 = dr::lerp(f010 - f000, f110 - f100, p_local.x());
        Float x1 = dr::lerp(f011 - f001, f111 - f101, p_local.x());
        Float dy = dr::lerp(x0, x1, p_local.z());
              x0 = dr::lerp(f001 - f000, f101 - f100, p_local.x());
              x1 = dr::lerp(f011 - f010, f111 - f110, p_local.x());
        Float dz = dr::lerp(x0, x1, p_local.y());

        return Vector3f(dx, dy, dz);
    }
```

rtabbara: Thanks! I don't think CUDA will do a texture gather for a 3d texture, so I think under the hood it will still be 8 individual lookups. However, imo, on the CPU-side what you've got looks cleaner to me so happy to give it a go.

src/shapes/sdfgrid.cpp
rtabbara: Thanks! Makes sense. I'll clean that up.

src/shapes/sdfgrid.cpp
```diff
@@ -494,57 +601,349 @@ class SDFGrid final : public SDF<Float, Spectrum> {
...
+
+        Transform<Point<FloatP, 4>> to_object = m_to_object.scalar();
+        Ray3fP ray = to_object.transform_affine(ray_);
+
+        auto shape = m_grid_texture.tensor().shape();
+
+        uint32_t voxel_index     = m_voxel_indices[prim_index];
+        ScalarVector3u voxel_pos = to_voxel_position(voxel_index);
+
+        // Find boxel AABB in object space
```
njroussel, line 636:
```suggestion
        // Find voxel AABB in object space
```

src/shapes/sdfgrid.cpp
```diff
@@ -562,70 +961,95 @@ class SDFGrid final : public SDF<Float, Spectrum> {
...
 
+        InputFloat f[6];
+        InputPoint3f query;
         Point3f voxel_size = 1.f / resolution;
-        Point3f p000 = Point3f(voxel_index) * voxel_size;
+        Point3f p000       = Point3f(voxel_index) * voxel_size;
 
         query = rescale_point(Point3f(p000[0] + voxel_size[0], p[1], p[2]));
-        m_grid_texture.eval(query, &f[0]);
+        m_grid_texture.eval(InputPoint3f(query), &f[0]);
```
njroussel, line 1040: I don't think all the `InputPoint3f` casts are necessary here, or are they? `rescale_point` should return a `InputPoint3f`.

src/shapes/sdfgrid.cpp
```diff
@@ -562,70 +961,95 @@ class SDFGrid final : public SDF<Float, Spectrum> {
...
+            jit_memcpy(JitBackend::CUDA, aabbs_ptr, aabbs,
+                       sizeof(BoundingBoxType) * count);
 
-        //TODO: async memcpy
-        void *aabbs_ptr = jit_malloc(AllocType::Device, sizeof(optix::BoundingBox3f) * count);
-        jit_memcpy(JitBackend::CUDA, aabbs_ptr, aabbs, sizeof(optix::BoundingBox3f) * count);
+            voxel_indices_ptr = (size_t *) jit_malloc(AllocType::Device,
+                                                      sizeof(size_t) * count);
+            jit_memcpy(JitBackend::CUDA, voxel_indices_ptr, voxel_indices,
+                       sizeof(size_t) * count);
```
njroussel, line 1013: I believe these can all be async operations: they will be added to the device queue of operations (CUDA stream). We shouldn't be reading/writing their values until the next kernel launch.

src/shapes/sdfgrid.cpp
njroussel: Indeed, `eval_fetch` is the way to go here. I wonder why I didn't do that initially...

njroussel (approved):
Mostly nitpicks/formatting stuff. Good job :rocket: !

Speierers:
🚀 great job!

src/render/shape.cpp
```diff
@@ -154,14 +155,16 @@ void embree_intersect_scalar(int* valid,
 #endif
         }
     } else {
-        if (dr::all(shape->ray_test(ray)))
+        if (dr::all(shape->ray_test(ray, true, primID)))
```
Speierers, line 158: Same here

include/mitsuba/render/shape.h
```diff
@@ -148,8 +151,11 @@ class MI_EXPORT_LIB Shape : public Object {
      *
      * \param ray
      *     The ray to be tested for an intersection
+     * 
+     * \param prim_index
+     *     Index of the primitive to be intersected
```
Speierers, line 156: Would be great to indicate what the default value `prim_index=0` do. Is this a special case?

src/render/shape.cpp
rtabbara: Thanks so much! I've fixed both these directly in the `sdf` branch

include/mitsuba/render/shape.h
rtabbara: Good point. I've added in the `sdf` branch a bit more info on the default behaviour and how for shape's that don't have multiple primitives this index is ignored. Hope that's fine.

PR #785, opened 2023-06-22
Fix epsilon issue for volpath integrators

dvicini:
The volpath integrators currently seem to have a small epsilon issue due to the use of `ref_interaction.spawn_ray` instead of `ref_interaction.spawn_ray_to`. The latter computes a better/more generous shadow epsilon. Therefore, replacing this call and the use of the previous shadow epsilon fixes the epsilon issue. 

For renderings showcasing the problem, see #678 (towards the bottom of the thread)

### DIFF
--- a/src/integrators/volpath.cpp
+++ b/src/integrators/volpath.cpp
@@ -339,7 +339,8 @@ public:
             return { emitter_val, ds };
         }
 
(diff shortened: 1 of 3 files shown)


Speierers:
LGTM

njroussel (approved):
Thanks!

PR #797, opened 2023-07-10
Disk: use tighter bounding box

bathal1:
## Description

Currently, the bounding box for a disk is computed as the bounding box of the square that inscribes it. One can however compute a tighter-fitting bounding box such that the disk is always tangent to some of its sides. See [here](https://iquilezles.org/articles/ellipses/) for the derivation.

Note that this implementation is already used for cylinders, that can be similarly bound by the bounding box of the bounding boxes of its endcaps.

## Testing

Tested in `scalar_rgb`, `cuda_ad_rgb` and `llvm_ad_rgb` for 1k random ellipses, resulting in a ~3% performance increase in CUDA mode, and 7% in LLVM mode.

### DIFF
--- a/src/shapes/disk.cpp
+++ b/src/shapes/disk.cpp
@@ -134,15 +134,17 @@ public:
 
 
     ScalarBoundingBox3f bbox() const override {
-        ScalarBoundingBox3f bbox;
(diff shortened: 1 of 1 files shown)


njroussel (approved):
LGTM! Thank you :smile:

PR #815, opened 2023-07-20
AOV Integrator: Add support to handle AD inner integrators 

rtabbara:
Motivated by issue reported in #668

## Description

Previously, the AOV integrator would attempt to perform AD back propagation to compute gradients for both the AOV and corresponding radiance channels. This would be irrespective of whether or not the inner integrators have their own custom implementation for computing backwards and forwards gradients (like prb)

The core issue is that `render_backward` or `render_forward` are always called by the AOV integrator to compute backward and forward gradients respectively, which use the default AD approaches. This then in-turn will call `render->sample` and thereafter calls the internal inner integrators' `sample` function. So we are forcibly performing gradient tracking for everything and the approach for computing gradients by the inner integrators is ignored entirely.

This PR modifies the AOV integrator to first explicitly split the render call into separate AOV and radiance images. This then allows us to split the `render_backward` and `render_forward` in two as well, and in particular allows us to call the inner integrators' corresponding `render_backward` and `render_forward` overrides (if they exist) with respect to the radiance. 

Note, this change requires defining the base default AD behaviour for `render_backward` and `render_forward` in cpp (where it previously was in Python) and we no longer rely on Python monkey patching.

## Testing

Added `test_aov.py` suite of unit tests


src/integrators/aov.cpp
```diff
@@ -184,6 +184,9 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
 
         if (m_aov_names.empty())
             Log(Warn, "No AOVs were specified!");
+
+        if (m_integrators.empty())
+            Throw("No child integrator specified!");
```
Speierers, line 189: It is sometimes useful to use the AOV integrator without a nested integrator, e.g. for quickly looking at shading normals or UV coordinates. Hence I wouldn't enfore the instantiation of a nested integrator.

src/films/hdrfilm.cpp
```diff
@@ -229,6 +229,19 @@ class HDRFilm final : public Film<Float, Spectrum> {
         props.mark_queried("banner"); // no banner in Mitsuba 3
     }
 
+    size_t target_base_channels_count() const override {
+        bool to_y = m_pixel_format == Bitmap::PixelFormat::Y 
+        || m_pixel_format == Bitmap::PixelFormat::YA;
```
Speierers, line 234: Minor: indentation

src/films/hdrfilm.cpp
```diff
@@ -229,6 +229,19 @@ class HDRFilm final : public Film<Float, Spectrum> {
...
+
+        /// Number of desired color components
+        uint32_t color_ch = to_y ? 1 : 3;
+
+        bool alpha = has_flag(m_flags, FilmFlags::Alpha);
+
+        /// Number of channels of the target tensor
+        return color_ch + (uint32_t)alpha;
```
Speierers, line 242: Minor
```suggestion
        return color_ch + (uint32_t) alpha;
```

src/integrators/aov.cpp
```diff
@@ -356,12 +425,98 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
...
+
+        dr::scatter(
+            dst.array(),
+            src.array(),
+            dst_values_idx);
+    }
+
+    /// Combine inner integrator images and AOVS image
+    TensorXf merge_channels(const std::vector<TensorXf>& inner_images, 
+                          const TensorXf& aovs_image) {
```
Speierers, line 465:
```suggestion
                            const TensorXf& aovs_image) {
```

src/integrators/aov.cpp
```diff
@@ -356,12 +425,98 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
...
+        }
+
+        // Load aovs image into combined
+        set_channels_slice(aovs_image, combined_image, channel_offset);
+
+        return combined_image;
+    }
+
+    /// Split up an image into image generated by inner-integrators and AOV image
+    std::pair<std::vector<TensorXf>, TensorXf> split_channels(size_t base_channel_count, const TensorXf& combined_image) const {
```
Speierers, line 494: Should those methods be placed in a different file, accessible to the rest of the codebase? Looks pretty generic functionalities to me.

src/render/integrator.cpp
```diff
@@ -41,6 +41,37 @@ Integrator<Float, Spectrum>::render(Scene *scene,
                   seed, spp, develop, evaluate);
 }
 
+MI_VARIANT typename Integrator<Float, Spectrum>::TensorXf
+Integrator<Float, Spectrum>::render_forward(Scene* scene,
+                                                    void* /*params*/,
+                                                    Sensor *sensor,
+                                                    uint32_t seed,
+                                                    uint32_t spp) {
```
Speierers, line 49: Indentation

src/render/integrator.cpp
```diff
@@ -41,6 +41,37 @@ Integrator<Float, Spectrum>::render(Scene *scene,
...
+    return TensorXf(dr::grad(image.array()), 3, image.shape().data());
+}
+
+MI_VARIANT void
+Integrator<Float, Spectrum>::render_backward(Scene* scene,
+                                                     void* /*params */,
+                                                     const TensorXf& grad_in,
+                                                     Sensor* sensor,
+                                                     uint32_t seed,
+                                                     uint32_t spp) {
```
Speierers, line 66: Ditto

src/render/python/integrator_v.cpp
```diff
@@ -77,6 +77,37 @@ MI_VARIANT class PySamplingIntegrator : public SamplingIntegrator<Float, Spectru
...
+                            uint32_t spp = 0) override {
+
+        py::gil_scoped_acquire gil;
+        py::function render_forward_override = py::get_override(this, "render_forward");
+
+        if (render_forward_override)
+            return render_forward_override(scene, params, sensor, seed, spp).template cast<TensorXf>();
+        else
+            return SamplingIntegrator::render_forward(scene, params, sensor, seed, spp);
+    }
```
njroussel, line 93: I'm unclear how this implementation would be different from just using the `PYBIND11_OVERRIDE` macro (this also applies to `render` and `render_backward`.

src/integrators/aov.cpp
njroussel, line 465:
```suggestion
                          const TensorXf& aovs_image) const {
```

src/integrators/aov.cpp
```diff
@@ -305,27 +300,101 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
...
+        for (auto& integrator : m_integrators) {
+            auto image = integrator->render(scene, sensor, seed, spp, develop, evaluate);
+            inner_images.push_back(image);
+        }
 
-                        if (ctr == 0)
-                            result = result_sub;
+        TensorXf aovs_image;
+        {
+            dr::suspend_grad<Float> scope(true);
```
njroussel, line 324: I think I'm missing something... If we're disabling gradients in this scope, how can the AOV fields be "attached" (to the AD graph) in `render_forward`/`render_backward`.

njroussel (changes requested):

src/integrators/aov.cpp
rtabbara: Good point, I've removed this constraint!

src/integrators/aov.cpp
```diff
@@ -305,27 +300,101 @@ class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
...
+            aovs_image = Base::render(scene, sensor, seed, spp, develop, evaluate);
 
-                        ctr++;
-                    }
-                    break;
-            }
+            // AOVs image above includes film target base channels as well so get slice
+            // just with AOVs
```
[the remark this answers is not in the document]
rtabbara: Yep!

src/integrators/aov.cpp
rtabbara: As discussed irl, while it doesn't impact render_forward/render_backward, it doesn't really need to be there so I've removed it.

src/render/python/integrator_v.cpp
rtabbara: Yep, I agree and have changed it

PR #828, opened 2023-07-27
Add `flip_tex_coords` option to `ply` plugin

### LINKED ISSUE #814: Add `flip_tex_coords` to .ply plugin
## Summary

It seems that texcoords mappings are flipped in the Y-axis after reading the mesh texcoords. Hence, it produced un-intuitive renders. It seems that `vertex_texcoords` are flipped in the Y-axis after loading the mesh. 

I can apply a `to_uv` transform to do the flip, but it would be good to have a similar functionality between .obj and .ply files, especially that .ply seems to be the recommended option. `flip_tex_coords` is defined once, while the alternative of `to_uv` transform requires me to instantiate the transform for every texture.

## Description

At the moment, my textures produce artefacts - black patches and incorrect coloration of the 3D object, due to incorrect Y-axis mapping.

Result vs Expected

Ply file header:

```
ply
format binary_little_endian 1.0
element vertex 159
property float x
property float y
property float z
property float nx
property float ny
property float nz
property float u
property float v
element face 240
property list uchar int vertex_indices
end_header
```

njroussel:
## Description

This PR adds the the option flip UV texture coordinates vertically on the `ply` plugin, similarly to the `obj` plugin.
By default, this option is set to false to guarantee backwards compatibility.

Fixes #814

## Testing

Some basic tests for this functionality were added.

### DIFF
--- a/src/render/tests/test_mesh.py
+++ b/src/render/tests/test_mesh.py
@@ -1007,4 +1007,47 @@ def test24_texture_attributes(variants_all_rgb):
     assert dr.allclose(mesh.eval_attribute('attribute_1', si), texture.eval(si))
 
 
+@fresolver_append_path
+def test25_flip_tex_coords_obj(variants_all_rgb, tmp_path):
+    mesh = mi.load_dict({
+        'type': 'obj',
+        'filename': 'resources/data/tests/obj/rectangle_uv.obj',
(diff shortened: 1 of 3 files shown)


Speierers (approved):
LGTM

Issue #676, opened 2023-04-18
BoundingSphere/Box Python binding ambiguity 


dvicini:
Hi,

I came across an edge case for `BoundingBox3f` and `BoundingSphere3f` in Python. If we first set the `llvm_ad_rgb` variant, changing to `scalar_rgb` will not recover the correct bindings for scalar mode (and vice versa). I came across this when running the Mitsuba test with `llvm_ad_rgb` set as a default variant. In this case, the bounding shape tests fail due to this incompatibility.

The problem is that the bounding shapes have a ray intersection function that is templated, and the binding differ between scalar mode and the scalar versions bound for the vectorized variants, e.g., `ScalarBoundingSphere3f`.

The following snippet illustrates the issue. The type `mi.scalar_rgb.BoundingSphere3f` will resolve to `mi.llvm_ad_rgb.ScalarBoundingSphere3f`, which will not accept scalar rays anymore.

```
import mitsuba as mi
mi.set_variant('llvm_ad_rgb')
mi.set_variant('scalar_rgb')

bsphere = mi.BoundingSphere3f()
print(type(bsphere))
bsphere.ray_intersect(mi.Ray3f([-5, 0, 0], [1, 0, 0]))

```
Or alternatively, first loading the `scalar_rgb` module:
```
import mitsuba as mi
mi.set_variant('scalar_rgb')
bsphere = mi.BoundingSphere3f()
mi.set_variant('llvm_ad_rgb')

bsphere = mi.ScalarBoundingSphere3f()
print(type(bsphere))
bsphere.ray_intersect(mi.Ray3f([-5, 0, 0], [1, 0, 0]))

```

Not sure what's the best solution here that doesn't require too many other changes. The Problem is inherently that we are trying to bind `BoundingSphere<Point<float, 3>>` twice, but with different arguments for the `ray_intersect` function. It seems that to properly solve this, the type of `Ray` would need to be a template parameter of `BoundingSphere` itself, rather than the function.

Speierers:
Nice catch! In that case could we bind a lambda that routes the function call to the right template at run time? (so that we don't bind a different function depending on the variant)

dvicini:
Interesting idea. We would still need some mechanism to deal with the fact that the bindings might be called in either order. Depending on the Python code, we might either bind `scalar_rgb.BoundingSphere3f` or `llvm_ad_rgb.ScalarBoundingSphere3f` first. 

[...]

PR #830, opened 2023-07-28
Properties: Support TensorXf and Transform3 

rtabbara:
## Description

1. Currently, plugin parameters that were of type `Transform3f` such as `to_uv` transforms, could only accept `mi.Properties` of type `Transform4f` and then we would have to extract down to the corresponding 3x3 transform. This change allows a user to directly set a `Transform3` e.g.
```
bsdf = mi.load_dict("bsdf": {
  "type": "diffuse",
  "reflectance": {
      "type": "checkerboard",
      "color0": rho0,
      "color1": rho1,
      "to_uv": mi.ScalarTransform3f.scale(2)
  },
})
```
Note, to maintain backwards compatibility, `props.get<ScalarTransform3f>` can still handle if the supplied transform is 4x4

2.  `mi.TensorXf` can now also be set in properties, meaning that in Python, plugin TensorXf parameters can be supplied during construction i.e.

```
my_plugin = mi.load_dict({"type : "plugin",  "some_tensor" : mi.TensorXf(array, shape) })
```
Additionally, all the implicit numpy, Tensorflow, PyTorch etc. conversions will still work in this context.

Note, that this change simply provides the necessary support within the Properties. It's up to the individual plugins themselves to modify the constructors to fetch and load the tensor instance

## Testing

Adding some additional tests in `test_properties.py`

### DIFF
--- a/include/mitsuba/render/volumegrid.h
+++ b/include/mitsuba/render/volumegrid.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <drjit/tensor.h>
+
 #include <mitsuba/core/bbox.h>
 #include <mitsuba/core/properties.h>
 #include <mitsuba/core/spectrum.h>
--- a/src/core/properties.cpp
+++ b/src/core/properties.cpp
@@ -9,6 +9,8 @@
 #include <sstream>
 #include <cstring>
 
+#include <drjit/tensor.h>
+
 #include <mitsuba/core/logger.h>
 #include <mitsuba/core/properties.h>
 #include <mitsuba/core/transform.h>
@@ -16,10 +18,12 @@
 
 NAMESPACE_BEGIN(mitsuba)
 
-using Float       = typename Properties::Float;
-using Color3f     = typename Properties::Color3f;
-using Array3f     = typename Properties::Array3f;
-using Transform4f = typename Properties::Transform4f;
+using Float         = typename Properties::Float;
+using Color3f       = typename Properties::Color3f;
+using Array3f       = typename Properties::Array3f;
+using Transform3f   = typename Properties::Transform3f;
+using Transform4f   = typename Properties::Transform4f;
+using TensorHandle  = typename Properties::TensorHandle;
 
 using VariantType = variant<
     bool,
@@ -27,7 +31,9 @@ using VariantType = variant<
     Float,
     Array3f,
     std::string,
+    Transform3f,
     Transform4f,
+    TensorHandle,
     Color3f,
     NamedReference,
     ref<Object>,
@@ -82,6 +88,23 @@ T get_impl(const Iterator &it) {
     return (T const &) it->second.data;
 }
 
+
+/**
+ * \brief Specialization to gracefully handle if user supplies either a 3x3 or 4x4 transform.
+ * Historically, we didn't directly support Transform3 properties so want to maintain 
+ * backwards compatibility
+ */
+template<>
+Transform3f get_impl<Transform3f, Transform4f>(const Iterator &it) {
+    if (!it->second.data.template is<Transform3f>() && !it->second.data.template is<Transform4f>())
+        Throw("The property \"%s\" has the wrong type (expected <%s> or <%s>, is <%s>)",
+              it->first, typeid(Transform3f).name(), typeid(Transform4f).name(), it->second.data.type().name());
+    it->second.queried = true;
+    if (it->second.data.template is<Transform4f>())
+        return ((Transform4f const &)it->second.data).extract();
+    return (Transform3f const &) it->second.data;
+}
+
 template <typename T>
 T get_routing(const Iterator &it) {
     if constexpr (dr::is_static_array_v<T>) {
@@ -92,6 +115,13 @@ T get_routing(const Iterator &it) {
         else
             return (T) get_impl<Array3f>(it);
     }
+
+    if constexpr (std::is_same_v<T, TensorHandle>)
+        return get_impl<TensorHandle>(it);
+
+    if constexpr (std::is_same_v<T, Transform<Point<float, 3>>> ||
+                  std::is_same_v<T, Transform<Point<double, 3>>>)
+        return (T) get_impl<Transform3f, Transform4f>(it);
 
     if constexpr (std::is_same_v<T, Transform<Point<float, 4>>> ||
                   std::is_same_v<T, Transform<Point<double, 4>>>)
@@ -171,7 +201,9 @@ T Properties::get(const std::string &name, const T &def_val) const {
 
 DEFINE_PROPERTY_SETTER(bool,         set_bool)
 DEFINE_PROPERTY_SETTER(int64_t,      set_long)
+DEFINE_PROPERTY_SETTER(Transform3f,  set_transform3f)
 DEFINE_PROPERTY_SETTER(Transform4f,  set_transform)
+DEFINE_PROPERTY_SETTER(TensorHandle, set_tensor_handle)
 DEFINE_PROPERTY_SETTER(Color3f,      set_color)
 DEFINE_PROPERTY_ACCESSOR(std::string,    string,  set_string,          string)
 DEFINE_PROPERTY_ACCESSOR(NamedReference, ref,     set_named_reference, named_reference)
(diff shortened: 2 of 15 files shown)


include/mitsuba/render/volumegrid.h
```diff
@@ -43,6 +45,7 @@ class MI_EXPORT_LIB VolumeGrid : public Object {
      */
     VolumeGrid(Stream *stream);
 
+
```
Speierers, line 48: Extra empty line

src/core/properties.cpp
```diff
@@ -171,7 +201,9 @@ T Properties::get(const std::string &name, const T &def_val) const {
 
 DEFINE_PROPERTY_SETTER(bool,         set_bool)
 DEFINE_PROPERTY_SETTER(int64_t,      set_long)
+DEFINE_PROPERTY_SETTER(Transform3f,  set_transform3f)
 DEFINE_PROPERTY_SETTER(Transform4f,  set_transform)
+DEFINE_PROPERTY_SETTER(TensorHandle,  set_tensor_handle)
```
Speierers, line 206: Minor

```suggestion
DEFINE_PROPERTY_SETTER(TensorHandle, set_tensor_handle)
```

src/core/python/properties_v.cpp
```diff
@@ -41,10 +47,14 @@ py::object properties_get(const Properties& p, const std::string &key) {
...
         return py::cast(p.get<dr::Array<PFloat, 3>>(key));
-    else if (type == Properties::Type::Transform)
+    else if (type == Properties::Type::Transform3f)
+        return py::cast(p.get<Transform<Point<PFloat, 3>>>(key));
+    else if (type == Properties::Type::Transform4f)
         return py::cast(p.get<Transform<Point<PFloat, 4>>>(key));
     // else if (type == Properties::Type::AnimatedTransform)
         // return py::cast(p.animated_transform(key));
+    else if (type == Properties::Type::Tensor)
+        return py::cast(*(TensorXf*)(p.get<TensorHandle>(key).get()));
```
Speierers, line 57: Could be worth printing a verbose error message here if the TensorHandle points to a tensor type of a different variant.

src/core/python/xml_v.cpp
```diff
@@ -362,6 +364,12 @@ void parse_dictionary(DictParseContext &ctx,
             continue;
         } catch (const pybind11::cast_error &) { }
 
+        // Try to cast to TensorXf
+        try {
+            props.set_tensor_handle(key, std::make_shared<TensorXf>(value.template cast<TensorXf>()));
```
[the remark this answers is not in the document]
Speierers: Another option would be to try to convert/cast it to the current variant.

src/shapes/sdfgrid.cpp
```diff
@@ -157,6 +157,15 @@ class SDFGrid final : public Shape<Float, Spectrum> {
             m_grid_texture = InputTexture3f(
                 InputTensorXf(vol_grid.data(), 4, shape), true, true,
                 dr::FilterMode::Linear, dr::WrapMode::Clamp);
+        } else if (props.has_property("grid")) {
+            TensorXf* tensor = reinterpret_cast<TensorXf*>(
```
Speierers, line 161: Could this be nicely packaged in a templated method of `Properties`?

src/textures/bitmap.cpp
```diff
@@ -164,103 +174,133 @@ class BitmapTexture final : public Texture<Float, Spectrum> {
...
+                else {
+                    const size_t pixel_count = tensor->shape(0) * tensor->shape(1);
+                    double mean = 0.0;
+                    ScalarFloat* ptr = tensor->data();
+                    for (size_t i = 0; i < pixel_count; ++i) {
+                        ScalarColor3f value = dr::load<ScalarColor3f>(ptr);
+                        mean += (double) luminance(value);
+                        ptr += 3;
+                    }
+                    m_mean = (ScalarFloat)(mean / pixel_count);
```
Speierers, line 204: We should probably support spectral modes as well here (same as what we do for the bitmap input).

Speierers (commented):
Great work! Looking forward to having this PR merged, it will be very useful!

EDIT: I just realized that this PR might actually still be WIP. If it is the case, maybe set it as draft PR so other people know it is still WIP. 😉

src/core/python/xml_v.cpp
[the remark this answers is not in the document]
njroussel: This _just works_:tm:

In `properites_v`: 
The [overload resolution order](https://pybind11.readthedocs.io/en/stable/advanced/functions.html?highlight=resolution%20order#overload-resolution-order) will attempt a conversion to `TensorXf` of the current `Properties` variant. No matter what the argument is, this is all handled by `tensor_init()` (even changes of variants).

In `xml_v.cpp`:
The caster will attempt to use the `TensorXf` constructor if the current type & variant does not match, in which case it will also resort to `tensor_init()`.

One problem that might exist is similar to https://github.com/mitsuba-renderer/mitsuba3/issues/676 I believe, where the `Properties` type itself doesn't update its bindings.

Admittedly, this is mostly pybind11 magic and took me a while to figure out why it works...

src/textures/bitmap.cpp
njroussel: The sRGB gamma conversion and the spectral up-sampling are scalar-only processes. I'd strongly prefer if it was somewhat explicit that a potential GPU->CPU conversion is necessary. Especially when the workaround is a straightforward one-liner: `mi.Bitmap(tensor_variable)`. 

In addition, I feel like it is clearer to a normal user that the `TensorXf` initialization is a fast path to setting data without anymodification ever.

src/shapes/sdfgrid.cpp
```diff
@@ -157,6 +157,14 @@ class SDFGrid final : public Shape<Float, Spectrum> {
             m_grid_texture = InputTexture3f(
                 InputTensorXf(vol_grid.data(), 4, shape), true, true,
                 dr::FilterMode::Linear, dr::WrapMode::Clamp);
+        } else if (props.has_property("grid")) {
+            TensorXf* tensor = props.tensor<TensorXf>("grid");
+            if (tensor->ndim() != 4)
```
njroussel, line 162: Let's add a check that the last dimension has size 1.

src/volumes/grid.cpp
```diff
@@ -175,82 +175,115 @@ class GridVolume final : public Volume<Float, Spectrum> {
...
+                if (!volume_grid)
+                    Throw("Property \"grid\" must be a VolumeGrid instance.");
+                res = volume_grid->size();
+                channel_count = volume_grid->channel_count();
+            } else if(props.has_property("data")) {
+                tensor = props.tensor<TensorXf>("data");
+                if (tensor->ndim() != 3 && tensor->ndim() != 4)
+                    Throw("Tensor->has %ul dimensions. Expected 3 or 4", tensor->ndim());
+                res = { (uint32_t) tensor->shape(2), (uint32_t) tensor->shape(1), (uint32_t) tensor->shape(0) };
+                channel_count = tensor->ndim() == 4 ? tensor->shape(3) : 1;
```
njroussel, line 205: I think we should have a check here that the number of channels is 1, 3 or 6.

njroussel (approved):
Nice :rocket: 
Two tiny comments

src/core/python/xml_v.cpp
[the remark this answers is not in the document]
Speierers: Awesome! Thanks for the explanations 😉

src/textures/bitmap.cpp
Speierers: I understand. But does this mean that in spectral mode, when passing a RGB image as a `TensorXf` or a `Bitmap`, it will produce different images?

njroussel: Here are all the cases:
```
(input type, color variant, raw?) => (expected semantic meaning of input data, processing step)

bitmap, rgb,      !raw  => (sRGB values or linear RGB, gamma correction)
tensor, rgb,      !raw  => not supported

bitmap, rgb,       raw  => (linear RGB, copied)
tensor, rgb,       raw  => (linear RGB, copied)

bitmap, spectral, !raw  => (sRGB values or linear RGB, spectral upsampling)
tensor, spectral, !raw  => not supported

bitmap, spectral,  raw  => (spectral upsampling coefficients, copied)
tensor, spectral,  raw  => (spectral upsampling coefficients, copied)
```
So if you have linear RGB values in a tensor, in a spectral variant, you will need to do:
`'bitmap': mi.Bitmap(tensor), 'raw': False`
And if you have sRGB values in a tensor, in a spectral variant, you will need to do:
```python
bitmap = mi.Bitmap(tensor); bitmap.set_srgb_gamma(True);
(...)
'bitmap': bitmap, 'raw': False
```

This is my current understanding of what will happen. Am I correct @rtabbara ?

I think I'm happy with this, any issues you see @Speierers? We're missing metadata about the tensor to be smart about sRGB conversions, there's no perfect solution IMHO.

Speierers: Ok, that looks good to me. Thanks for the detailed table.

As long as we don't get different results, I am happy with the current solution. Here it seems that it either matches what `Bitmap` would do, or it is explicitly not supported. 👍

Speierers:
🚀

PR #854, opened 2023-08-17
Check that dict keys do not contains . character

Speierers:
It is important to prevent users from using the `.` character in scene object's names, as this character is already used as path separator (e.g. in `mi.traverse()`). This PR adds a simple check for this in the `mi.load_dict()` routine, suggesting the user to use `_` instead.


njroussel:
Great suggestion, thank you!

PR #857, opened 2023-08-17
Add Scene.sensors_dr() and bindings

Speierers:
Similar to `Scene::shapes_dr()` and `Scene::emitters_dr()`, this PR adds `Scene::sensors_dr()`.


njroussel (approved):
I'm surprised we didn't realize this earlier!
Thanks.

PR #858, opened 2023-08-17
Bitmap plugin don't store unnecessary bitmap object

Speierers:
This PR removes the unnecessary storage of the original `Bitmap` object passed to the constructor of the `bitmap` texture plugin.


njroussel (approved):
I've added a tiny fix to your initial change to the `bitmap` texture. We had a similar case in the `gridvolume` plugin and I applied a similar fix. 
Thank you

PR #859, opened 2023-08-17
Fix BSDF pure trampolines

Speierers:
An implementation of the `BSDF::eval_diffuse_reflectance` method should be optional when using the BSDF trampoline. This PR fixes this.


njroussel (approved):
Thank you!

PR #860, opened 2023-08-17
Add support for hide_emitters in python integrators

Speierers:
This PR adds support to hide directly visible emitters to all Python integrators. This uses the already existing `hide_emitters` property parsed in all integrator's constructor.


njroussel (approved):
:+1:

PR #863, opened 2023-08-17
Fix prbvolpath NEE throughput

Speierers:
This PR fixes a issue in `prbvolpath` when the phase function sampling routine returns a weight different than `1.0`. 

Currently, the path throughput used in NEE contains the phase function sampling weight which is wrong. We never noticed this issue beforehand as most phase functions return a sampling weight of one. In order to fix this issue, we propose to move the phase function sampling after performing NEE.

Thanks to [elite-sheep](https://github.com/elite-sheep) for tracking down this bug.


njroussel (approved):
Indeed, nice catch!! Thanks.

(Ping @mcrescas - most likely a fix you'd want to be aware of)

PR #883, opened 2023-08-29
Envmap: do not assume alignment in RGB mode

merlinND:
## Description

The `envmap` plugin used to always use 4 channels regardless of the color mode. When updating the luminance map in `parameters_changed()`, aligned stores and loads were used, which was fine.

However, the envmap can now use 3 channel data in RGB mode (https://github.com/mitsuba-renderer/mitsuba3/commit/820c38e575b5e9e45fa89e046871c88a20181ef1), so the alignment assumption no longer holds.

This PR switches to unaligned loads and stores when the envmap data uses 3 channels.

## Testing

Added a simple test which triggers `Envmap::parameters_changed()`. When unconditionally assuming alignment, this test crashes in RGB modes.

## Checklist


njroussel (approved):
Thanks for the fix!

(I'm surprised we never ran into this ...)

Issue #461, opened 2022-12-20
The bounding box obtained from `scene.bbox()` donot change with the transformation


swordigo1995:
## Summary

Hi, I found the bounding box of scene obtained by `scene.bbox()` is not changed when I modify the transformation of object in the scene. 

## System configuration

System information:

OS: Ubuntu 20.04
CPU: Intel Core i7-7700
GPU: NVIDIA RTX 2080Ti
Python version: 3.9.13
CUDA version: 11.6
NVidia driver: 515.57

Dr.Jit version: 0.3.2
Mitsuba version: 3.1.1

## Steps to reproduce
```
import drjit as dr
import mitsuba as mi
mi.set_variant('cuda_ad_rgb')

scene = mi.load_file("../scenes/simple.xml")
params = mi.traverse(scene)

iter_num = 5
for i in range(iter_num):
    # Translate the teapot a little bit
    V = dr.unravel(mi.Point3f, params['teapot.vertex_positions'])
    V.z += 0.5
    params['teapot.vertex_positions'] = dr.ravel(V)
    params.update();
    modified_image = mi.render(scene, spp=256)
    print(scene.bbox())
```

njroussel:
Hi @swordigo1995 

Are there any other objects than the teapot in the scene?

swordigo1995:
No. Here is the scene xml:
```
<scene version='3.0.0'>
    <default name="spp" value="256"/>
    <default name="res" value="256"/>

    <integrator type='direct'/>

    <sensor type="perspective" id="sensor">
        <transform name="to_world">
            <lookat target="0.0, 0.0, 1.25"
                    origin="0.0, -12.0, 5.0"
                    up    ="0.0, 0.0, 1.0"/>
        </transform>
        <sampler type="independent">
            <integer name="sample_count" value="$spp"/>
        </sampler>
        <film type="hdrfilm">
            <rfilter type="box"/>
            <integer name="width"  value="$res"/>
            <integer name="height" value="$res"/>
        </film>
    </sensor>

    <shape type="ply" id="teapot">
        <string name="filename" value="meshes/teapot.ply"/>
        <bsdf type="diffuse">
            <rgb name="reflectance" value="0.9 0.9 0.0"/>
        </bsdf>
    </shape>

</scene>
```
I also found the 'bbox' of 'PLYMesh' is changed correctly if I `print(scene)`.

njroussel:
[...]

PR #886, opened 2023-08-29
Scene: Update bounding box when relevant params changed

rtabbara:
## Description

Motivated by #461, in the event that the underlying accel structure needs to be recomputed, we similarly need to recompute corresponding scene bounding box 

## Testing

Added a small test in `test_scene`


njroussel (approved):
All good :rocket:

PR #887, opened 2023-08-30
irradiancemeter: check if m_shape is set

danielwujecki:
When the irradiancemeter is instantiated on it's own instead of as a child of a shape, mitsuba runs into a segmentation fault. Introducing a simple check in the constructor makes sure that the scene format is valid and the irradiancemeter is used as a child of the shape plugin.


njroussel:
Hi @danielwujecki 

What you've pointed out is indeed a problem.

However, your suggested fix is wrong. The `m_shape` member will only be set by the enclosing shape `Shape::initialize()` and this is only called once the Shape plugin is constructed. Because the scene construction is a bottom-up process, the `irradiancemeter` is instantiated before the `shape` that encloses it. It is therefore impossible to have this check in the `irradiancemeter`'s constructor.

I'll close this issue for now, we can re-open it if you find an other workaround.

Also, **when creating a PR, please fill out the PR template. It save us a lot of time. Thank you.**

PR #888, opened 2023-08-30
Add OptiX debug and validation flags option in CMake

rtabbara:


njroussel:
I kept cherry-picking these commits without ever merging them ....

Let's go :rocket:

Issue #900, opened 2023-09-12
AOV plugin returns an exception


joeylitalien:
## Summary

AOV plugin is broken.

## System configuration

System information:

  OS: Ubuntu 20.04.6 LTS
  CPU: Intel(R) Core(TM) i7-9800X CPU @ 3.80GHz
  GPU: NVIDIA TITAN V
  Python: 3.10.13 (main, Aug 25 2023, 13:20:03) [GCC 9.4.0]
  NVidia driver: 510.39.01
  CUDA: 11.6.55
  LLVM: 12.0.0

  Dr.Jit: 0.4.3
  Mitsuba: 3.4.0
     Is custom build? False
     Compiled with: GNU 10.2.1
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        llvm_ad_rgb

## Description

The AOV plugin returns a critical exception in all RGB variants.

## Steps to reproduce

1. Render a simple scene with `aov` integrator as stated in the docs:
```
<scene version="2.1.0">
    <default name="spp" value="16"/>
    <default name="resx" value="512"/>
    <default name="resy" value="512"/>

    <integrator type="aov">
        <string name="aovs" value="d:depth,nn:sh_normal"/>
        <integrator type="path" name="my_image"/>
    </integrator>

    <sensor type="perspective">
        <float name="fov" value="30"/>
        <transform name="to_world">
            <lookat origin="5, 5, 0" target="0, 0, 0" up="0, 1, 0"/>
        </transform>
        <sampler type="independent">
            <integer name="sample_count" value="$spp"/>
        </sampler>
        <film type="hdrfilm">
            <integer name="width" value="$resx"/>
            <integer name="height" value="$resy"/>
        </film>
    </sensor>

    <emitter type="constant">
        <rgb name="radiance" value="1.0"/>
    </emitter>

    <shape type="sphere">
        <bsdf type="diffuse">
            <rgb name="reflectance" value="0.5"/>
        </bsdf>
    </shape>
</scene>
```

Somehow this throws the error: `Caught a critical exception: Tensor::shape(2): out of bounds!`.

rtabbara:
Hi @joeylitalien ,

Thanks for reporting this. We've tracked down the issue and a PR should hopefully be in soon

rtabbara:
Fixed in #901 and part of 3.4.1 release

PR #901, opened 2023-09-13
AOV Integrator: Fixing up primal render

rtabbara:
Withing the `AOVIntegrator`, partitioning of render calls between AOV and base channels was implemented in #815 . This was motivated by the need to separate how gradients are computed between AOV channels which simply use autodiff and the inner integrators that may invariably have their own handling of `render_forward` and `render_backwards`

However, writing to an underlying film is tightly coupled to an integrator's render call. In particular, the integrator is in effect responsible for initialising the film prior to rendering. With respect to the AOV Integrator, this is problematic because while we want to split our render calls between AOVs and inner integrators for the purposes of differentiation, we concurrently require a single render call that will write both AOV and base channels into our film. This PR is an attempt to address the latter problem.

### DIFF
--- a/src/integrators/aov.cpp
+++ b/src/integrators/aov.cpp
@@ -103,11 +103,34 @@ public:
         dUVdx,
         dUVdy,
         PrimIndex,
-        ShapeIndex
+        ShapeIndex,
+        IntegratorRGBA
     };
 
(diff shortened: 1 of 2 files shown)


njroussel:
Fixes #900 and #921

njroussel (approved):
LGTM :rocket:

PR #902, opened 2023-09-14
Add dict support in Chi2 adapters

Speierers:
This PR adds support for Python dictionary to instantiate plugins in Chi2 adapters.


njroussel (approved):
Thanks, this is a nice addition :+1:

PR #903, opened 2023-09-14
Python bindinds improvements

Speierers:
This PR brings the following improvement to the Python bindings of Mitsuba:

a. Add bindings for the `lookup_ior` routine
b. Binds aliases for the `ComplexNf` Dr.Jit types
c. Fix bindings of the `TraverseCallback::put_parameter` function in order to be able to call `put_parameter` from Python on a `TraverseCallback` class defined in C++.
d. Fix the `Integrator::render_forward` and `Integrator::render_backward` bindings to properly handle the `params` arguments.

### DIFF
--- a/src/core/python/object_v.cpp
+++ b/src/core/python/object_v.cpp
@@ -4,6 +4,35 @@
 #include <drjit/dynamic.h>
 #include <drjit/tensor.h>
 
+using Caster = py::object(*)(mitsuba::Object *);
+extern Caster cast_object;
+
+// Trampoline for derived types implemented in Python
+class PyTraversalCallback : public TraversalCallback {
+public:
+    void put_parameter_impl(const std::string &name, void *ptr,
+                            uint32_t flags, const std::type_info &type) override {
+        py::gil_scoped_acquire gil;
+        py::function overload = py::get_overload(this, "put_parameter");
+
+        if (overload)
+            overload(name, ptr, flags, (void *) &type);
+        else
+            Throw("TraversalCallback doesn't overload the method \"put_parameter\"");
+    }
+
+    void put_object(const std::string &name, Object *obj,
+                    uint32_t flags) override {
+        py::gil_scoped_acquire gil;
+        py::function overload = py::get_overload(this, "put_object");
+
+        if (overload)
+            overload(name, cast_object(obj), flags);
+        else
+            Throw("TraversalCallback doesn't overload the method \"put_object\"");
+    }
+};
+
 /// Defines a list of types that plugins can expose as tweakable/differentiable parameters
 #define APPLY_FOR_EACH(T) \
     T(Float32); T(Float64); T(Int32); T(UInt32); T(DynamicBuffer<Float32>);    \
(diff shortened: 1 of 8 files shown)


src/render/python/integrator_v.cpp
```diff
@@ -216,7 +216,15 @@ MI_VARIANT class PyADIntegrator : public CppADIntegrator<Float, Spectrum> {
...
-        PYBIND11_OVERRIDE(TensorXf, Base, render_forward, scene, params, sensor, seed, spp);
+        py::gil_scoped_acquire gil;
+
+        py::function func = py::get_override(this, "render_forward");
+        if (func) {
+            auto o = func(scene, *((py::object *) params), sensor, seed, spp);
+            return o.template cast<TensorXf>();
+        }
+
+        return Base::render_forward(scene, params, sensor, seed, spp);
```
njroussel, line 227: I'm curious about this change. 

My guess is that the macro's behavior would not do the explicit casting and de-referencing of `params`. But then, what was happening?

src/core/python/object_v.cpp
```diff
@@ -25,18 +54,32 @@
...
     if (strcmp(type.name(), typeid(T).name()) == 0) {                          \
         *((T *) ptr) = py::cast<T>(value);                                     \
         return;                                                                \
     }
 
+/// Macro to iterate over types that can be passed to put_parameter_impl
+#define PUT_PARAMETER_IMPL_T(T)                                                \
+    if (py::isinstance<T>(value)) {                                            \
+        T v = py::cast<T>(value);                                              \
+        self->put_parameter_impl(name, &v, flags, typeid(T));                  \
```
njroussel, line 74: I'm trying to understand the necessity of `PublicistTraversalCallback`.

Why can't we use `put_parameter` here? We can template it on `T`, can't we ?

njroussel (commented):
Looks good, I need a couple clarifications though  :+1:

src/core/python/object_v.cpp
Speierers: We need this because `put_parameter` is a protected method.

src/render/python/integrator_v.cpp
Speierers: Without this explicit cast, on the Python side the `params` variable wouldn't be recognise as having the right type (e.g. `SceneParameters`).

src/core/python/object_v.cpp
njroussel: I think you mean `put_parameter_impl` is protected.

That's exactly my point, `put_parameter` is public so you can just use that instead and avoid the publicist.

Speierers: Sorry for the confusion. The issue here with `put_parameter` is that it required the `typeid` of `value` which is only known at run-time. Hence we need this macro to generate a big switch statement that checks for all the possible types.

If there is another way around (likely more elegant than this) I would be happy to take it!

njroussel:
All good, thanks for the clarifications :+1:

PR #904, opened 2023-09-14
Various fixes in integrators/common.py

Speierers:
This PR fixes various minor issues in the `common.py` file.

a. Rename `aovs()` method into `aovs_names()` to match the C++ name
b. Remove unnecessary storage of the rendered image in `self.primal_image`
c. Various other improvements

### DIFF
--- a/src/python/python/ad/integrators/common.py
+++ b/src/python/python/ad/integrators/common.py
@@ -42,9 +42,6 @@ class ADIntegrator(mi.CppADIntegrator):
         # Warn about potential bias due to shapes entering/leaving the frame
         self.sample_border_warning = True
 
-    def aovs(self):
-        return []
-
     def to_string(self):
         return f'{type(self).__name__}[max_depth = {self.max_depth},' \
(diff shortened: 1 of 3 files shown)


src/python/python/ad/integrators/common.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, props = mi.Properties()):
         # Warn about potential bias due to shapes entering/leaving the frame
         self.sample_border_warning = True
 
-    def aovs(self):
+    def aovs_names(self):
```
njroussel, line 45: The method in C++ is actually called `aov_names` and we already have bindings for it, so we don't even need to re-define it here I believe.

njroussel (changes requested):

src/python/python/ad/integrators/common.py
Speierers: Ouach nice catch!

njroussel (approved):
Nice, thanks for the changes :rocket:

PR #905, opened 2023-09-15
Fix normalmap

tomix1024:
## Description

The shading frame in the `normalmap` BSDF was not computed correctly.
The shading frame-local normal was used in conjunction with the world-space `dp_du` to construct the perturbed shading frame.
The shading frame in turn was interpreted as the "shading frame in world space", as well as the relative transformation between the original shading frame and the perturbed one...

## Testing

`resources/data/tests/scenes/git/test_normalmap_roughconductor_path.xml` has been added in my fork of the [mitsuba-data repository](https://github.com/tomix1024/mitsuba-data) to verify the changes.
**(But this probably does not really count here, does it?)**
The improved correctness relative to the previous implementation can be seen if the normal-mapped rendering is compared to a not normal-mapped rendering.

I have added `src/bsdfs/tests/test_normalmap.py`, that verifies that sampling from a normal-mapped BSDF is the same as sampling from the nested BSDF with an appropriately rotated surface interaction.

## Checklist

### DIFF
--- a/src/bsdfs/normalmap.cpp
+++ b/src/bsdfs/normalmap.cpp
@@ -118,9 +118,10 @@ public:
                                              const Point2f &sample2,
                                              Mask active) const override {
         // Sample nested BSDF with perturbed shading frame
+        auto [ perturbed_frame_wrt_si, perturbed_frame_wrt_world ] = frame(si, active);
         SurfaceInteraction3f perturbed_si(si);
-        perturbed_si.sh_frame = frame(si, active);
-        perturbed_si.wi = perturbed_si.to_local(si.wi);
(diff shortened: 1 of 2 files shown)


tomix1024:
Please have a look if the tests are sufficient as they are now...
The interesting case occurs only if an anisotropically scattering BSDF is nested inside of the normalmap, so it is difficult to have some "analytic" ground truth solution to test against.

njroussel:
This got lost in the stack of other issues/PRs. Sorry, about that. The PR is correct, the incorrect frame was being computed for the new local shading. I can't update this PR directly, so I've opened #1479 with your original commits and some necessary changes.
Thank you

PR #917, opened 2023-09-25
Fix missing field assignment in Bitmap copy constructor

dvicini:
The Bitmap's move and copy constructors were missing the assignment for the m_premultiplied_alpha field.


njroussel (approved):
LGTM !

PR #930, opened 2023-10-06
PRBVolpath: Re-evaluate phase function sampling weight in a differentiable way.

bathal1:
The current implementation of `prbvolpath` samples the phase function of a volume in a completely detached way. As a consequence, gradients are missing when computing derivatives of an objective w.r.t. phase function parameters.

In the reproducer below, we compute the derivative of a dummy loss w.r.t. the `g` parameter of a Henyey-Greenstein phase function and compare to finite differences. The current implementation produces the following (wrong) result:

```
FD derivative: -0.06458275765180588
AD derivative: -0.6339979767799377
```

In order to get valid gradient, we need to re-evaluate the phase function value with gradient tracking enabled, and then divide by the detached pdf, like we already do for BSDF sampling. This then gives the expected derivatives:

```
FD derivative: -0.06458361446857452
AD derivative: -0.06470370292663574
```

Here is the reproducer. Note that we need to run the computation several times and average the result, as running it with a high number of samples runs into numerical issues when using reverse-mode AD.

```python
import mitsuba as mi
mi.set_variant('cuda_ad_rgb')
from mitsuba.scalar_rgb import Transform4f as T
import drjit as dr

scene_dict = {
    'type': 'scene',
    'integrator': {
        'type': 'prbvolpath',
        'max_depth': 4,
        'rr_depth': 4,
    },
    'sensor': {
        'type': 'perspective',
        'fov': 15,
        'to_world': T.look_at(origin=[0, 0, 3], target=[0, 0, 0], up=[0, 1, 0]),
        'film': {
            'type': 'hdrfilm',
            'width': 64,
            'height': 64,
        },
    },
    'sphere': {
        'type': 'sphere',
        'to_world': T.rotate((0, 0, 1), 180),
        'bsdf': {
            'type': 'null',
        },
        'interior': {
            'type': 'homogeneous',
            'phase': {
                'type': 'hg',
                'g': 0.8,
            },
        }
    },
    'emitter': {
        'type': 'constant',
    }
}


a = mi.Float(0.8)
key = 'sphere.interior_medium.phase_function.g'
eps = 1e-3

grad_fd = 0
N = 64
spp = 256
params = mi.traverse(scene)
for i in range(N):
    params[key] = a + eps
    params.update()
    img1 = mi.render(scene, params=params, seed=i, spp=16*spp)
    loss1 = dr.mean(dr.sqr(img1))

    params[key] = a - eps
    params.update()
    img0 = mi.render(scene, params=params, seed=i, spp=16*spp)
    loss0 = dr.mean(dr.sqr(img0))

    grad_fd += (loss1 - loss0) / (2 * eps)
del scene

print(f"FD derivative: {grad_fd[0]/N}")


grad_ad = 0
params = mi.traverse(scene)
for i in range(N):
    a = mi.Float(0.8)
    dr.enable_grad(a)
    dr.set_grad(a, 0)
    params[key] = a
    params.update()
    img = mi.render(scene, params=params, seed=i, spp=spp)
    loss = dr.mean(dr.sqr(img))
    dr.backward(loss)
    grad_ad += dr.grad(a)
del scene

print(f"AD derivative: {grad_ad[0]/N}")

mi.Bitmap(img).write("primal.exr")
```

### DIFF
--- a/src/python/python/ad/integrators/prbvolpath.py
+++ b/src/python/python/ad/integrators/prbvolpath.py
@@ -278,10 +278,20 @@ class PRBVolpathIntegrator(RBIntegrator):
                                                                sampler.next_2d(act_medium_scatter),
                                                                act_medium_scatter)
                 act_medium_scatter &= phase_pdf > 0.0
+
+                # Re evaluate the phase function value in an attached manner
+                phase_eval, _ = phase.eval_pdf(phase_ctx, mei, wo, act_medium_scatter)
+                if not is_primal and dr.grad_enabled(phase_eval):
+                    Lo = phase_eval * dr.detach(dr.select(active, L / dr.maximum(1e-8, phase_eval), 0.0))
+                    if mode == dr.ADMode.Backward:
+                        dr.backward_from(δL * Lo)
+                    else:
+                        δL += dr.forward_to(Lo)
+
+                throughput[act_medium_scatter] *= phase_weight
                 ray[act_medium_scatter] = mei.spawn_ray(wo)
                 needs_intersection |= act_medium_scatter
                 last_scatter_direction_pdf[act_medium_scatter] = phase_pdf
-                throughput[act_medium_scatter] *= phase_weight
 
                 # ------------------------ BSDF sampling -----------------------
 


src/python/python/ad/integrators/prbvolpath.py
```diff
@@ -278,10 +278,20 @@ def sample(self,
                                                                sampler.next_2d(act_medium_scatter),
                                                                act_medium_scatter)
                 act_medium_scatter &= phase_pdf > 0.0
+
+                # Re evaluate the phase function value in an attached manner
+                phase_eval, _ = phase.eval_pdf(phase_ctx, mei, wo, act_medium_scatter)
+                if not is_primal and dr.grad_enabled(phase_eval):
+                    Lo = phase_eval * dr.detach(dr.select(active, L / dr.maximum(1e-8, phase_eval), 0.0))
```
njroussel, line 285: Just to double-check, this would be "_more_" correct:
```suggestion
                    Lo = phase_eval * dr.detach(dr.select(act_medium_scatter, L / dr.maximum(1e-8, phase_eval), 0.0))
```
Right?
Both are functionally and numerically equivalent (because `phase_eval` will be a detached `0` if `act_medium_scatter` is `False` and `active` is True`), but as a reader it seems weird that the masking is different..

njroussel (approved):
This looks good, thanks :rocket: !

One tiny comment.

src/python/python/ad/integrators/prbvolpath.py
bathal1: Yes, you're right.

PR #940, opened 2023-10-12
Avoid copies to host in 1D distributions

njroussel:
## Description

This PR modifies `DiscreteDistribution` and `ContinuousDistribution` (both 1D) such that they no longer perform device to host copies in order to compute their CDF. These computations can now entirely be done on device thanks to the changes introduced in https://github.com/mitsuba-renderer/drjit-core/commit/91201eff4a543e6ccadcfd4d7b0d7505bfdee7d7.

## Testing

No new tests were added. These changes are purely for performance, the classes should functionally remain the same.


njroussel:
(Converted to draft - I think I can actually handle the `IrregularDistribution`)

include/mitsuba/core/distr_1d.h
```diff
@@ -209,7 +214,26 @@ template <typename Value> struct DiscreteDistribution {
...
+            Throw("DiscreteDistribution: entries must be non-negative!");
+        if (!dr::any(m_pmf > 0.f))
+            Throw("DiscreteDistribution: no probability mass found!");
+
+        m_cdf = dr::prefix_sum(m_pmf, false);
+        m_valid = ScalarVector2u(0, m_pmf.size() - 1);
+        m_sum = dr::gather<Float>(m_cdf, m_pmf.size() - 1);
+        m_normalization = 1.0 / m_sum;
+
+        dr::make_opaque(m_pmf, m_cdf, m_sum, m_normalization);
```
wjakob, line 233: make_opaque should not be needed here. It would only be non-opaque when the entire PMF is uniform, which is probably a nice special case where a specialized kernel is helpful.

include/mitsuba/core/distr_1d.h
```diff
@@ -294,17 +318,15 @@ template <typename Value> struct ContinuousDistribution {
...
         : m_pdf(dr::load<FloatStorage>(values, size)),
           m_range(range) {
-        compute_cdf(values, size);
+        compute_cdf_scalar(values, size);
     }
 
     /// Update the internal state. Must be invoked when changing the pdf.
     void update() {
         if constexpr (dr::is_jit_v<Float>) {
-            FloatStorage temp = dr::migrate(m_pdf, AllocType::Host);
```
wjakob, line 303: so nice!

include/mitsuba/core/distr_1d.h
```diff
@@ -392,32 +414,40 @@ template <typename Value> struct ContinuousDistribution {
...
         MI_MASK_ARGUMENT(active);
 
-        value *= m_integral;
+        sample *= m_integral;
 
         Index index = dr::binary_search<Index>(
             m_valid.x(), m_valid.y(),
             [&](Index index) DRJIT_INLINE_LAMBDA {
-                return dr::gather<Value>(m_cdf, index, active) < value;
+                //return dr::gather<Value>(m_cdf, index, active) < value;
```
wjakob, line 431: leftover comment

include/mitsuba/core/distr_1d.h
```diff
@@ -392,32 +414,40 @@ template <typename Value> struct ContinuousDistribution {
...
+                Value value = dr::gather<Value>(m_cdf, index, active);
+                if constexpr (!dr::is_jit_v<Float>) {
+                    return value < sample;
+                } else {
+                    // `m_valid` did not trim the leading/traling zeros in the PMF
+                    return ((value < sample) || (dr::eq(value, 0))) &&
+                           dr::neq(value, m_integral);
```
wjakob, line 438: Here, there is a check against `m_integral`. Above, it was against `m_normalization`. Is that right?

include/mitsuba/core/distr_1d.h
```diff
@@ -472,7 +502,41 @@ template <typename Value> struct ContinuousDistribution {
     }
 
 private:
-    void compute_cdf(const ScalarFloat *pdf, size_t size) {
+    void compute_cdf() {
+        static_assert(dr::is_jit_v<Float>,
```
wjakob, line 506: I think this is technically not legal. You can static_assert on something that is a template parameter of a function. But when it's a template parameter of the class, then that function is in a sense already instantiated when the class is instantiated (since it's not a template function). It's probably just laziness in the compiler that this static assert isn't triggered.

include/mitsuba/core/distr_1d.h
```diff
@@ -472,7 +502,41 @@ template <typename Value> struct ContinuousDistribution {
...
+        if (!(m_range.x() < m_range.y()))
+            Throw("ContinuousDistribution: invalid range!");
+        if (!dr::all(m_pdf >= 0.f))
+            Throw("DiscreteDistribution: entries must be non-negative!");
+        if (!dr::any(m_pdf > 0.f))
+            Throw("DiscreteDistribution: no probability mass found!");
+
+        uint32_t size = m_pdf.size() - 1;
+        m_interval_size_scalar = (m_range.y() - m_range.x()) / size;
+        auto index_1_to_n = dr::arange<dr::uint32_array_t<Float>>(1, size + 1);
```
wjakob, line 520: Give type a name using `using = ` and then avoid autos? (e.g. `UInt32`)

include/mitsuba/core/distr_1d.h
```diff
@@ -472,7 +502,41 @@ template <typename Value> struct ContinuousDistribution {
...
+
+        m_cdf =
+            m_interval_size_scalar *
+            (dr::gather<Float>(dr::prefix_sum(m_pdf, false), index_1_to_n) -
```
wjakob, line 524: I didn't quite understand the logic here -- you're doing a prefix sum and then throwing away the first element. Isn't that what an `exclusive` prefix` sum does (with `true` second argument)

include/mitsuba/core/distr_1d.h
```diff
@@ -472,7 +502,41 @@ template <typename Value> struct ContinuousDistribution {
...
+             0.5 * dr::gather<Float>(m_pdf, index_1_to_n));
+
+        m_valid = ScalarVector2u(0, size);
+        m_integral = dr::gather<Float>(m_cdf, size - 1);
+        m_normalization = 1.0 / m_integral;
+        m_interval_size = m_interval_size_scalar;
+        m_inv_interval_size = 1.f / m_interval_size;
+        m_max = dr::slice(dr::max(m_pdf));
+
+        dr::make_opaque(m_pdf, m_cdf, m_integral, m_normalization,
```
wjakob, line 535: Ditto for opaqueness.

wjakob (commented):
Thank you @njroussel. A few comments on this first version inline below.

include/mitsuba/core/distr_1d.h
```diff
@@ -209,7 +214,26 @@ template <typename Value> struct DiscreteDistribution {
...
+                      "This method should only be used in JIT variants!");
+
+        if (m_pmf.empty())
+            Throw("DiscreteDistribution: empty distribution!");
+        if (!dr::all(m_pmf >= 0.f))
+            Throw("DiscreteDistribution: entries must be non-negative!");
+        if (!dr::any(m_pmf > 0.f))
+            Throw("DiscreteDistribution: no probability mass found!");
+
+        m_cdf = dr::prefix_sum(m_pmf, false);
```
ziyi-zhang, line 228: Previously I did this but it complains about `m_pmf` attached to the AD graph. 
So i detached it before feeding it to prefix

wjakob: The operation is differentiable in the next version of Dr.Jit, but the one on nanobind master raises an exception when given differentiable input to be on the safe side. Detaching is a reasonable workaround for now, and I think we usually in any case don't want to differentiate with respect to the PMF. @ziyi-zhang , it might be worth investigating why `m_pmf` is differentiable in the first place.

njroussel: I haven't added any explicit `dr::detach` statements, just in case someone eventually wants to differentiate through the PMF/PDF. I don't mind adding them, if you feel like the risk of accidentally differentiating w.r.t. the distribution outweighs this _future_ feature.

include/mitsuba/core/distr_1d.h
```diff
@@ -392,32 +411,39 @@ template <typename Value> struct ContinuousDistribution {
...
 
         Index index = dr::binary_search<Index>(
             m_valid.x(), m_valid.y(),
             [&](Index index) DRJIT_INLINE_LAMBDA {
-                return dr::gather<Value>(m_cdf, index, active) < value;
+                Value value = dr::gather<Value>(m_cdf, index, active);
+                if constexpr (!dr::is_jit_v<Float>) {
+                    return value < sample;
+                } else {
+                    // `m_valid` did not trim the leading/traling zeros in the PDF
```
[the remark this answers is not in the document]
njroussel: That is correct. In JIT variants, `m_valid` is always exactly equal to the full range `(0, CDF size - 1)`.

I didn't think that I could find the first/last occurrence of a non-zero value in the PDF/PMF in JIT-mode efficiently. But actually, now that I think of it again, I could figure out the correct `m_valid` range with a binary search on the CDF in `compute_cdf`.
Is this what you're getting at?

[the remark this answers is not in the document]
wjakob: Neat, but then it would synchronize I think. It was more of a comment about updating the comment.

PR #947, opened 2023-10-18
Ensure correct rotation of backfacing retarders

dvicini:
Hi,

I think there is a small bug in the retarder implementation. If a retarder is rotated, the rotation's sign should be flipped if the retarder is viewed from the reverse direction.

This is necessary to ensure the following physical property: If a retarder is rotated 45 degrees, it will convert linearly vertically polarized light to left circularly polarized light in one direction. In the opposite direction, it should be doing the reverse, which is currently not the case.

Intuitively, it also makes sense that a clockwise rotation on one side would be a counterclockwise rotation on the other side.

The fix is easy, but I would like to have confirmation from someone else to make sure this is really the right thing here.

### DIFF
--- a/src/bsdfs/retarder.cpp
+++ b/src/bsdfs/retarder.cpp
@@ -111,13 +111,15 @@ public:
             UnpolarizedSpectrum delta = dr::deg_to_rad(m_delta->eval(si, active));
 
             // Approximate angle-of-incidence behaviour with a cosine falloff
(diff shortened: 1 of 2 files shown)


tizian:
Ah nice catch, that indeed makes more sense! Maybe you can add an additional comment to explain this additional `sign(cos_theta)` term. IMO there can never be too many of these to explain the subtleties of polarization. 😄

PR #953, opened 2023-10-24
Fix version checks in BlenderMesh plugin

Speierers:
This PR fixes the Blender version checks in the `BlenderMesh` plugin. More specifically, the current checks assume there is no major version greater than 3, which isn't the case anymore with the "soon-to-be-released" Blender v4.0.

This patch reuses the `Version` class implemented in `xml.cpp`, now moved to `util.h`, to perform the checks.

### DIFF
--- a/include/mitsuba/core/util.h
+++ b/include/mitsuba/core/util.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <mitsuba/mitsuba.h>
+#include <mitsuba/core/string.h>
+#include <mitsuba/core/logger.h>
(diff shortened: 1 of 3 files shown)


bathal1:
That's a nice improvement, thanks @Speierers ! We could take the opportunity to throw an error if the blender version is not supported, but maybe it's best to do that on the exporter side.

Speierers:
> maybe it's best to do that on the exporter side.

Yes probably better to do it on the add-on side.

bathal1 (approved):

PR #1013, opened 2023-12-15
Add support for AOVs to Python AD integrators.

dvicini:
## Description
 PR to add AOV support to Python AD integrators. The motivation for this is to be able to output arbitrary auxiliary values from Python AD integrators. This can be useful to add additional supervision, e.g., to segmentation masks for inverse rendering.

## Checklist

### DIFF
--- a/src/python/python/ad/integrators/common.py
+++ b/src/python/python/ad/integrators/common.py
@@ -74,13 +74,14 @@ class ADIntegrator(mi.CppADIntegrator):
             ray, weight, pos = self.sample_rays(scene, sensor, sampler)
 
             # Launch the Monte Carlo sampling process in primal mode
-            L, valid, _ = self.sample(
+            L, valid, aovs, _ = self.sample(
                 mode=dr.ADMode.Primal,
                 scene=scene,
                 sampler=sampler,
                 ray=ray,
                 depth=mi.UInt32(0),
                 δL=None,
+                δaovs=None,
                 state_in=None,
                 active=mi.Bool(True)
             )
@@ -97,6 +98,7 @@ class ADIntegrator(mi.CppADIntegrator):
                 value=L * weight,
                 weight=1.0,
                 alpha=dr.select(valid, mi.Float(1), mi.Float(0)),
+                aovs=aovs,
                 wavelengths=ray.wavelengths
             )
 
@@ -131,7 +133,7 @@ class ADIntegrator(mi.CppADIntegrator):
             ray, weight, pos = self.sample_rays(scene, sensor, sampler)
 
             with dr.resume_grad():
-                L, valid, _ = self.sample(
+                L, valid, aovs, _ = self.sample(
                     mode=dr.ADMode.Forward,
                     scene=scene,
                     sampler=sampler,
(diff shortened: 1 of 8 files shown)


dvicini:
I think this is working. The failing tests are due to one of the tutorials needing to be updated. I will make a PR to the tutorials repo as well, but maybe let's first make sure that this PR is in good shape and the interface is the one we agree on.

src/python/python/ad/integrators/common.py
```diff
@@ -721,45 +742,53 @@ def splatting_and_backward_gradient_image(value: mi.Spectrum,
                 with dr.suspend_grad(pos, ray, weight):
                     L = dr.full(mi.Spectrum, 1.0, dr.width(ray))
                     dr.enable_grad(L)
-
+                    aovs = []
+                    for _ in self.aov_names():
+                        aov = dr.full(mi.Float, 1.0, dr.width(ray))
```
Speierers, line 747:
```suggestion
                        aov = dr.ones(mi.Float, dr.width(ray))
```

src/python/python/ad/integrators/common.py
```diff
@@ -721,45 +742,53 @@ def splatting_and_backward_gradient_image(value: mi.Spectrum,
...
                     splatting_and_backward_gradient_image(
                         value=L * weight,
                         weight=1.0,
-                        alpha=1.0
+                        alpha=1.0,
+                        aovs=[aov * weight for aov in aovs]
                     )
 
                     δL = dr.grad(L)
+                    δaovs = [dr.grad(aov) for aov in aovs]
```
Speierers, line 758:
```suggestion
                    δaovs = dr.grad(aovs)
```

src/render/python/integrator_v.cpp
```diff
@@ -247,28 +247,28 @@ MI_VARIANT class PyADIntegrator : public CppADIntegrator<Float, Spectrum> {
...
+                "δaovs"_a=py::none(),
                 "state_in"_a=py::none(),
                 "active"_a=active
             );
             // Third output is the ADIntegrator's state, which we ignore here
-            auto [spec, mask, _] =
+            auto [spec, mask, aovs_, _] =
                 sample_override(**kwargs).template cast<PyReturn>();
-
+            std::copy(aovs_.begin(), aovs_.end(), aovs);
```
Speierers, line 271: Does this actually work? I suppose at this point `aovs_` is already converted to a `std::vector<Float>`?

In my case I was casting the variables explicitly:

```c++
    // Convert AOVs from python object to float array if needed
    if (py::isinstance<py::list>(outputs)) {
        auto list = outputs.template cast<py::list>();
        for (size_t i = 0; i < list.size(); ++i)
            aovs[i] = list[i].template cast<Float>();
    }
 ```

Speierers (commented):
Those changes look good to me 🚀 

You already addressed what I mentioned in our discussion on Slack in `integrator_v.cpp` regarding the broken trampoline for Python AOV integrators.

src/render/python/integrator_v.cpp
dvicini: Yes, I think so. I followed the already existing code from the PySamplingIntegrator. We are explicitly casting the return type to an std::vector. I think the code you wrote is marginally more efficient, as it does not require the temporary std::vector, but I would favor readability & simplicity here.

dvicini:
Thanks for the feedback @Speierers ! I incorporated your suggestions

njroussel (approved):
Thanks for all the work!

It looked good, I pushed a new version with some minor changes. I'll take over the remaining changes if there any failing tests :+1:

dvicini:
Thanks, appreciate it!

PR #1022, opened 2023-12-21
Projective cont.

ziyi-zhang:
This PR keeps small changes to projective integrator related stuff.

Changed:
- Turn the flag `sqrt_scale_mass` into a float in [0, 1] to transform sample contribution with x'=x^(1-alpha). This gives more flexibility to control the suppression of outliers.

Question:
- Why `octree_scatter_inc` is by default false?

(re-gen docstring before merging)


njroussel (approved):
> Why octree_scatter_inc is by default false?

I must have been debugging right before the release. I'll fix this right away.

> (re-gen docstring before merging)

I'll do it once the PR is merged, it will be easier. 


Thanks

PR #1052, opened 2024-01-26
Fix SGGX Python bindings

dvicini:
Hi,

The SGGX Python bindings currently expect a dr::Array<Float, 6>, which we don't actually bind to Python. I am not sure if this was always the case, or maybe the Vector6f bindings were removed. In either case, we probably do not want to bind Vector6f.

This PR changes the SGGX functions to expect an std::array<Float, 6> instead. It also adds a Python-side test of the sampling and PDF evaluation, which ensures that those bindings will not break easily.

### DIFF
--- a/src/render/python/microflake_v.cpp
+++ b/src/render/python/microflake_v.cpp
@@ -4,21 +4,37 @@
 MI_PY_EXPORT(MicroflakeDistribution) {
     MI_PY_IMPORT_TYPES()
 
-    m.def("sggx_sample",
-          py::overload_cast<const Frame3f&,
-                            const Point2f&,
-                            const dr::Array<Float, 6>&>(sggx_sample<Float>),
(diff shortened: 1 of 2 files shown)


wjakob:
My suggestion would be that we keep this PR open tentatively but do not merge it into the mainline repository. The upcoming nanobind version will make it super-easy to add extra Dr.Jit types in other projects, and there are some benefits with keeping the native Dr.Jit static arrays that are considered "traversable" for AD and similar purposes (whereas std::array is not).

dvicini:
Okay I see, that's fine with me. Let's wait for Nanobind then and add the missing bindings with that change. I agree that having the native Dr.Jit arrays is much more convenient than dealing with Python lists of Float.

I will leave the PR open and adapt it once the Nanobind support is there.

njroussel:
Replaced by #1276

PR #1057, opened 2024-02-02
Fix use-after-free and memory leak

merlinND:
Fixes two bugs found using AddressSanitizer, using the methodology described in: https://github.com/mitsuba-renderer/drjit-core/pull/78.

### DIFF
--- a/include/mitsuba/core/argparser.h
+++ b/include/mitsuba/core/argparser.h
@@ -89,6 +89,12 @@ public:
     };
 
 public:
+
+    ~ArgParser() {
+        for (Arg *arg : m_args)
+            delete arg;
+    }
+
     /**
      * \brief Register a new argument with the given prefix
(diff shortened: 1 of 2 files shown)


include/mitsuba/core/argparser.h
```diff
@@ -89,6 +89,12 @@ class MI_EXPORT_LIB ArgParser {
     };
 
 public:
+
+    ~ArgParser() {
+        for (Arg *arg : m_args)
+            delete arg;
```
merlinND, line 95: This assumes that `ArgParser::add()` takes ownership of the created `Arg`, which I think is fair.

wjakob: Yep, that's how it is intended.

wjakob:
Thank you, Merlin!

PR #1069, opened 2024-02-12
Fix Properties::SortKey operator for large integers

Speierers:
The current implementation of the key comparison operator in `Properties` assumes that integers found in keys fit in a 32 bits integer. When working with large scenes exported from Blender, it is often the case for object names to contain larger integers. This results in Mitsuba complaining about missing or duplicated keys.

This patch fixes this issue by performing a proper comparison for integers that fit in 64 bits, and otherwise falling back onto the string comparison.

### DIFF
--- a/src/core/properties.cpp
+++ b/src/core/properties.cpp
@@ -8,6 +8,7 @@
 #include <map>
 #include <sstream>
 #include <cstring>
+#include <climits>
 
 #include <drjit/tensor.h>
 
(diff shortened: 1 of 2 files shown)


njroussel (approved):
LGTM, you're just missing an `#include <climits>` statement for GCC.

Speierers:
Should be ready to go now 🚀

Issue #703, opened 2023-04-28
How to avoid memory illegal access over time (memory leak)?


ShnitzelKiller:
## Summary

Generating and rendering many thousands of scenes in succession eventually leads to a segfault, seemingly due to a memory leak.

## System configuration

Rocky Linux release 9.1 (64-bit)

System information:
```
   CPU: Intel(R) Xeon(R) Silver 4110 CPU @ 2.10GHz
  GPU: NVIDIA TITAN Xp
NVIDIA TITAN Xp
NVIDIA TITAN Xp
NVIDIA TITAN Xp
  Python: 3.10.9 | packaged by conda-forge | (main, Feb  2 2023, 20:20:04) [GCC                     11.3.0]
  NVidia driver: 515.65.01
  LLVM: -1.-1.-1

  Dr.Jit: 0.4.1
  Mitsuba: 3.2.1
     Is custom build? False
     Compiled with: GNU 10.2.1
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        llvm_ad_rgb
```
## Description

I am generating many scenes and rendering them to images in my python code. The code works fine, but after a while, I inevitably get the error:
```
Critical Dr.Jit compiler failure: cuda_check(): API error 0700 (CUDA_ERROR_ILLEGAL_ADDRESS): "an illegal memory access was encountered" in /project/ext/drjit-core/src/init.cpp:448.
Aborted (core dumped)
```
I assume that this is due to a memory leak, since I always get through about the same number of scenes before this error happens again and I have to restart over again.

My question is, is there currently something I can do, such as best practices, to prevent this from happening? Some way of explicitly clearing some kind of cache that fills up over time? Or some other thing that python garbage collection does not handle in the case of Mitsuba?

## Steps to reproduce
In a loop, call a subroutine which loads a simple scene and renders that scene thousands of times, and it will eventually hit a segfault.

shallowtoil:
Hi @ShnitzelKiller,

I've encountered the same problem. Have you figured out how to tackle this?

njroussel:
Hi!

We have encountered this ourselves in recent experiments. This most likely is a bug deep within DrJit. We have found that using the `CUDA_LAUNCH_BLOCKING` flag (set as an environment variable) seems to drastically reduce the likelihood of this segfault happening.

@ShnitzelKiller @shallowtoil  could you elaborate on your specifc setups?

Something super basic like:
```python
import mitsuba as mi
[...]

PR #1072, opened 2024-02-14
Update compiling.rst

dnakath:
Added compilation instructions for Ubuntu 22.04 as a small documentation improvement.

## Description

I added compilation instructions for Ubuntu 22.04, while trying to keep the original structure of the instructions, as briefly discussed [here](https://github.com/mitsuba-renderer/mitsuba3/issues/703#issuecomment-1937872954).


njroussel:
Hi @dnakath 

I appreciate the thought. But I don't think we'll merge this.
The currently listed environment is a description of our CI.  I don't believe that there is anything wrong with the package versions you suggest, we are simply not testing against it. There's a slim risk that your suggested versions are no longer compatible with Mitsuba down the line and we wouldn't become aware of it immediately.

PR #1073, opened 2024-02-15
Fix reference counting of OptiX shader binding table

Speierers:
This patch attempts to fix a reference counting issue with OptiX's shader binding tables shared between scenes.

More specifically, this issue arises when a mesh is shared in between two scenes, with the `eval_parameterization()` being used on that mesh. When the first scene is deleted, it will free the shader binding table used in `eval_parameterization()` which we might want to use with the second scene. See the pseudo code below to illustrate this example:

```python
shape = ...

scene1 = mi.load_dict({ 'type': 'scene', 'shape': shape })

a = scene1.ray_intersect(...)
b = shape.eval_parameterization()
dr.eval(a, b)

del scene1

scene2 = mi.load_dict({ 'type': 'scene', 'shape': shape })

a = scene2.ray_intersect(...)
b = shape.eval_parameterization()
dr.eval(a, b)
```

Unfortunately this patch doesn't solve all possible issues. When a Mesh is shared in between multiple scenes, it should ideally have a parameterization nested scene per parent scene, to make sure `eval_parameterization` and `scene->ray_tracing()` can take place within the same megakernel.

Any thoughts?

### DIFF
--- a/src/render/scene_optix.inl
+++ b/src/render/scene_optix.inl
@@ -30,12 +30,11 @@ struct OptixSceneState {
     OptixAccelData accel;
     OptixTraversableHandle ias_handle = 0ull;
     struct InstanceData {
-        void* buffer = nullptr;             // Device-visible storage for IAS 
-        void* inputs = nullptr;             // Device-visible storage for OptixInstance array
+        void* buffer = nullptr;  // Device-visible storage for IAS
+        void* inputs = nullptr;  // Device-visible storage for OptixInstance array
     } ias_data;
     size_t config_index;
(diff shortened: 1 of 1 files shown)


wjakob:
This change looks fine to me.

njroussel:
LGTM.
FYI I've found a few other leaks and reference counting issues in the OptiX scene setup when writing the new nanobind bindings. I'll try to backport them.

I'm not sure how we could ever sove that last case you mentionned.

dvicini:
Having those fixes to leaks & reference counting backported would be appreciated @njroussel 👍

PR #1083, opened 2024-02-21
Fix Properties::SortKey trailing zeros

Speierers:
This PR is fixing another edge case in `Properties::SortKey`: when the key's numerical suffixes compile to the same value, but have a different number of trailing zeros, the operator should fallback to the string comparison.

For instance, currently the following two keys can be used together in a `props`: `key_001` and `key_01`.

I also added a unit test that covers this edge case.

### DIFF
--- a/src/core/properties.cpp
+++ b/src/core/properties.cpp
@@ -65,7 +65,7 @@ struct SortKey {
             if (a_end == (a.c_str() + a.size()) &&
                 b_end == (b.c_str() + b.size()) &&
                 l1 != LLONG_MAX && l2 != LLONG_MAX &&
-                !((l1 == l2) && (a.size() != b.size()))) // catch trailing zeros case (e.g. 001 vs 01))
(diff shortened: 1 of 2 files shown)


src/core/properties.cpp
```diff
@@ -64,7 +64,8 @@ struct SortKey {
             long long l2 = std::strtoll(b_ptr, &b_end, 10);
             if (a_end == (a.c_str() + a.size()) &&
                 b_end == (b.c_str() + b.size()) &&
-                l1 != LLONG_MAX && l2 != LLONG_MAX)
+                l1 != LLONG_MAX && l2 != LLONG_MAX &&
+                !((l1 == l2) && (a.size() != b.size()))) // catch trailing zeros case (e.g. 001 vs 01))
```
njroussel, line 68:
```suggestion
                !((l1 == l2) && (a.size() != b.size()))) // catch leading zeros case (e.g. 001 vs 01))
```

njroussel (approved):
LGTM

njroussel:
Thank you @Speierers !

PR #1130, opened 2024-04-10
Docs: update recommended Clang version

merlinND:
Hi,

@jhoydis pointed out recently that the current build guide's instructions:

```bash
sudo apt install clang-10 libc++-10-dev libc++abi-10-dev cmake ninja-build
```

no longer work on e.g. Ubuntu 22.04 (and most likely more recent versions as well), because the `libc++-10-dev` and `libc++abi-10-dev` packages no longer exist.

I understand that Clang 10 is used in the CI and it makes sense to recommend the same compiler version that is tested, but since it's not working anymore, would switching to Clang 15 in the docs be okay?


njroussel:
Done in https://github.com/mitsuba-renderer/mitsuba3/commit/7bce33d92df92bf72b3c310b39d10c45fafb0a6e

PR #1172, opened 2024-05-14
[bug fix] issue #1168 by adding eval_null_transmission to trampoline class

### LINKED ISSUE #1168: Python BSDF inconsistent with C++ BSDF: eval_null_transmission not working at all
I have a problem with bsdfs defined in Python: eval_null_transmission seems to never be used. I am working on a BSDF for a transmissive material in Python, but it currently is a light trap, except for when there is an environment emitter. I implemented the same thing as a C++ plugin, and it works without issue. The same functionality implemented in Python does not work though. I assume that eval_null_transmission is not properly called in the Python version. When I try deliberately causing a runtime error in that method, no error occurs. I think this might be similar to #1162  but I can not figure out, how the signature of my function differs from the API reference.

Here is a shortened version of my Python code. The expected behavior would be for the object to be completely transmissive like the null bsdf. The only difference is an additional reflection component (the corresponding code was edited out by me because it behaves exactly as expected and is not relevant to the problem).
The actual behavior of the Python BSDF is that the surface it is applied to becomes completely black as long as there is no environment emitter (when a diffuse object is inside the transmissive object, some white dots of Monte Carlo noise appear). When there is an environment emitter, the material appears completely transmissive as it should be. 
```py
class PrototypeBSDF(mi.BSDF):
    def __init__(self, props):
        mi.BSDF.__init__(self, props)
        reflection_flags   = mi.BSDFFlags.DeltaReflection   | mi.BSDFFlags.FrontSide | mi.BSDFFlags.BackSide
        transmission_flags = mi.BSDFFlags.Null |mi.BSDFFlags.FrontSide | mi.BSDFFlags.BackSide
        
        self.m_components  = [reflection_flags, transmission_flags]
        self.m_flags = reflection_flags | transmission_flags
        
    def sample(self, ctx, si, sample1, sample2, active):
        bs = mi.BSDFSample3f()
        bs.wo = -si.wi
        bs.eta = 1
        bs.pdf = 1
        bs.sampled_component = 1
        bs.sampled_type = mi.UInt32(+mi.BSDFFlags.DeltaTransmission)

        weight = mi.Color3f(1)
        return (bs, weight)
    
    def eval(self, ctx, si, wo, active):
        return 0.0

    def pdf(self, ctx, si, wo, active):
        return 0.0
   
    def eval_null_transmission(self, si, active=True):
        return mi.Color3f(1)
    
mi.register_bsdf("prototype", lambda props: PrototypeBSDF(props))
```
For comparison, my C++ code (except for the additional component functionally identical to the null bsdf). The C++ code works perfectly fine. I can not figure out, how this code differs from the Python implementation, except that eval_null_transmission in the Python code is probably never called

```c++
PrototpyeBSDF(const Properties &props) : Base(props) {
        m_components.push_back(BSDFFlags::DeltaReflection | BSDFFlags::FrontSide |
                               BSDFFlags::BackSide);
        m_components.push_back(BSDFFlags::Null | BSDFFlags::FrontSide | BSDFFlags::BackSide);
        m_flags = m_components[0] | m_components[1];
        dr::set_attr(this, "flags", m_flags);
    }

    std::pair<BSDFSample3f, Spectrum> sample(const BSDFContext &ctx,
                                             const SurfaceInteraction3f &si,
                                             Float sample1,
                                             const Point2f & /* sample2 */,
                                             Mask active) const override {
        MI_MASKED_FUNCTION(ProfilerPhase::BSDFSample, active);

        BSDFSample3f bs = dr::zeros<BSDFSample3f>();
        bs.wo = -si.wi;
        bs.eta = 1.f;
        bs.pdf = 1.f;
        bs.sampled_component = UInt32(1);
        bs.sampled_type = UInt32(+BSDFFlags::DeltaTransmission);
        UnpolarizedSpectrum weight = 1;
        return { bs, depolarizer<Spectrum>(weight) & active };
    }

    Spectrum eval(const BSDFContext & /* ctx */, const SurfaceInteraction3f & si,
                  const Vector3f & wo, Mask active) const override {
        return 0.f;
    }

    Float pdf(const BSDFContext & /* ctx */, const SurfaceInteraction3f & /* si */,
              const Vector3f & /* wo */, Mask /* active */) const override {
        return 0.f;
    }

    Spectrum eval_null_transmission(const SurfaceInteraction3f & si,
                                    Mask active) const override {
        UnpolarizedSpectrum value = 1;
        return depolarizer<Spectrum>(value);
    }
```

As a workaround I will try to rewrite my Python code o feature only the reflection lobe and nest it in the working C++ implementation.

## System configuration

Mitsuba and Drjit are custom compiled, I only added additional plugins in C++ and have not changed any preexisting functionality. All of Mitsuba 3 is up to date, the issue also exists with a build based on a ~5 months old version of MItsuba 3.

DanielPSchenk:
[bug fix] for issue #1168

## Description

This small contribution fixes an issue I had as suggested by @merlinND.

Fixes #1168 

## Testing

I verified that eval_null_transmission is called as expected after the fix and that my python BSDF (see #1168) behaves exactly the same as the reference C++ plugin.

## Checklist

- [x] I have commented my code: not necessary
- [x] I have made corresponding changes to the documentation: not necessary
- [ ] I have added tests that prove my fix is effective or that my feature works: I verified that it fixes the problem I had

### DIFF
--- a/src/render/python/bsdf_v.cpp
+++ b/src/render/python/bsdf_v.cpp
@@ -65,6 +65,11 @@ public:
         PYBIND11_OVERRIDE(Spectrum, BSDF, eval_diffuse_reflectance, si, active);
     }
 
+    Spectrum eval_null_transmission(const SurfaceInteraction3f &si,
+                                      Mask active) const override {
+        PYBIND11_OVERRIDE(Spectrum, BSDF, eval_null_transmission, si, active);
+    }
+
     Mask has_attribute(const std::string &name, Mask active) const override {
         PYBIND11_OVERRIDE(Mask, BSDF, has_attribute, name, active);
     }


njroussel (approved):
Hi @DanielPSchenk 

Sorry, the CI ran out of memory. This is not an issue with your changes, we can ignore it.
Thank you for the fix!

PR #1222, opened 2024-07-10
MergeShape: add to registry upon construction

merlinND:
## Description

Without this change, any inclusion of `merge` in a scene would crash at the `jit_registry_remove(this)` call in the destructor of `MergeShape`, due to the missing `jit_registry_put()` call.

I checked all other plugins that should be included in the registry,. Since they all call their `Base` constructor, they shouldn't have this issue.

## Testing

Add basic unit tests, since this plugin was never instantiated in the test suite.

## Checklist


njroussel (approved):
LGTM, thanks for the extra tests too!

Issue #1254, opened 2024-08-06
Cannot set `mi.BSDF` on `Properties`


dvicini:
The following code used to work to instantiate a new Mesh, but now fails:
```
import mitsuba as mi 

mi.set_variant('llvm_ad_rgb')

props = mi.Properties()
props['bsdf'] = mi.load_dict({'type': 'diffuse'})
mi.Mesh(props)
```
with the error:

```
TypeError: __setitem__(): incompatible function arguments. The following argument types are supported:
    1. __setitem__(self, arg0: str, arg1: float, /) -> None
    2. __setitem__(self, arg0: str, arg1: bool, /) -> None
    3. __setitem__(self, arg0: str, arg1: int, /) -> None
    4. __setitem__(self, arg0: str, arg1: str, /) -> None
    5. __setitem__(self, arg0: str, arg1: mitsuba.ScalarColor3f) -> None
    6. __setitem__(self, arg0: str, arg1: mitsuba.ScalarColor3d) -> None
    7. __setitem__(self, arg0: str, arg1: drjit.scalar.Array3f64, /) -> None
    8. __setitem__(self, arg0: str, arg1: mitsuba.ScalarTransform3d, /) -> None
    9. __setitem__(self, arg0: str, arg1: mitsuba.ScalarTransform4d, /) -> None
    10. __setitem__(self, arg0: str, arg1: ref<mitsuba::Object>, /) -> None
    11. __setitem__(self, arg0: str, arg1: drjit.llvm.ad.TensorXf, /) -> None

Invoked with types: mitsuba.Properties, str, mitsuba.Object
```
I tried explicitly converting to `mi.Object`, but that didn't help.

PR #1256, opened 2024-08-07
Fix props `ref<Object>` set/get support

rtabbara:


dvicini:
Adding the extra import seems to indeed fix #1254, thanks!

njroussel (approved):
All good, thanks 

Fixes #1254

PR #1260, opened 2024-08-09
More Mitsuba Python stubs improvements

rtabbara:
- Going through `pyright` output to identity some legitimate issues with the bindings, primarily with repeated bindings.
- For the variant submodules, including them in the top-level namespace so that the relative imports in the stubs are variant agnostic.
- Attempt to express `Spectrum` and `UnpolarizedSpectrum` in a variant-agnostic way that still provides some specificity as a templated `dr.ArrayBase` type
- Generating stubs for native and Python submodules


njroussel (approved):
I just rebased onto `master` and updated the default variants. Looks good :D

PR #1264, opened 2024-08-12
Fix Python ScopedThreadEnvironment

### LINKED ISSUE #1263: Python ScopedThreadEnvironment not working
The ScopedThreadEnvironment seems broken in Python currently. Reproducer:

```
import mitsuba as mi
mi.set_variant("llvm_ad_rgb")

env = mi.ThreadEnvironment()
with mi.ScopedSetThreadEnvironment(env):
  mi.Log(mi.LogLevel.Warn, "Log from a thread env")
```
This prints:
```
2024-08-12 14:58:04 WARN  main  [Thread] Destructor called while thread 'main' was still running
2024-08-12 14:58:04 WARN  py0   [thread_env2.py:6] <module>: Log from a thread env
Traceback (most recent call last):
  File ".../mitsuba3/build/thread_env2.py", line 5, in <module>
    with mi.ScopedSetThreadEnvironment(env):
TypeError: __exit__(): incompatible function arguments. The following argument types are supported:
    1. __exit__(self, arg0: object, arg1: object, arg2: object, /) -> None

Invoked with types: mitsuba.ScopedSetThreadEnvironment, NoneType, NoneType, NoneType
```

Likely just an issue in the bindings, but not immediately obvious to me what causes this. Aside of that, it seems that there are currently no tests exercising the ScopedThreadEnvironment in Python.

dvicini:
This fixes #1263. The issue was the missing "_a.none()" annotation for Nanobind. The naming of the parameters to `__exit__` follows what Dr.Jit itself uses. I also added a new test that covers that at least using a thread environment does not fail.

### DIFF
--- a/src/core/python/thread.cpp
+++ b/src/core/python/thread.cpp
@@ -112,6 +112,7 @@ MI_PY_EXPORT(Thread) {
                                             D(ScopedSetThreadEnvironment))
         .def(nb::init<const ThreadEnvironment &>())
         .def("__enter__", &PyScopedSetThreadEnvironment::enter)
-        .def("__exit__", &PyScopedSetThreadEnvironment::exit);
+        .def("__exit__", &PyScopedSetThreadEnvironment::exit,
+             "exc_type"_a.none(), "exc_val"_a.none(), "exc_tb"_a.none());
 }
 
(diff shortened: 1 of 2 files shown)


njroussel:
Hi @dvicini 

This looks good!

(Ignore the CI for now, looks like it's stuck/broken because of recent update of nanobind. I'll retrigger it manually, once I figure what's going on.)

PR #1266, opened 2024-08-13
SceneParameter: fix update parameters

rtabbara:
For custom Python plugins that expose modifiable parameters, updating these values through the use of `mi.traverse` required the use of `assign` to enable updating of the underlying cpp data. i.e. we can propagate the modifications to the actual plugin parameters.

For this particular case, we instead rely on nanobind's `nb::inst_replace_copy` function to perform the same functionality.

The behaviour updating parameters of cpp plugins remains unchanged.

### DIFF
--- a/src/core/python/object_v.cpp
+++ b/src/core/python/object_v.cpp
@@ -101,6 +101,18 @@ MI_PY_EXPORT(Object) {
         },
         "ptr"_a, "type"_a, "value"_a);
 
+    m.def(
+        "set_property",
+        [](nb::handle dst, nb::handle src) {
+            nb::handle h = dst.type();
+            if (nb::type_check(h)) {
+                nb::inst_replace_copy(dst, h(src));
+                return;
+            }
+            Throw("set_property(): Target property type isn't a nanobind type!");
+        },
+        "dst"_a, "src"_a);
+
     if constexpr (dr::is_array_v<ObjectPtr>) {
         dr::ArrayBinding b;
         auto object_ptr = dr::bind_array_t<ObjectPtr>(b, m, "ObjectPtr");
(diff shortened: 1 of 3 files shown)


dvicini:
Amazing - I was about to re-raise this. I tried the fix and it works for me.

It would be great to also add a unit test.

rtabbara:
Good point! I've updated the `test_util.py` tests to check whether changes to params actual affect the corresponding plugin variables for both cpp plugins and custom Python plugins.

dvicini:
Looks great!

src/core/python/object_v.cpp
```diff
@@ -101,6 +101,18 @@ MI_PY_EXPORT(Object) {
...
 
+    m.def(
+        "set_property",
+        [](nb::handle dst, nb::handle src) {
+            nb::handle h = dst.type();
+            if (nb::type_check(h)) {
+                nb::inst_replace_copy(dst, h(src));
+                return;
+            }
+            Throw("set_property(): Target property type isn't nanobind type!");
```
njroussel, line 112:
```suggestion
            Throw("set_property(): Target property type isn't a nanobind type!");
```

src/python/python/util.py
```diff
@@ -73,9 +73,9 @@ def __setitem__(self, key: str, value):
 
         if value_type is None:
             try:
-                cur.assign(value)
-            except AttributeError as e:
-                if "has no attribute 'assign'" in str(e):
+                self.set_property(cur, value)
+            except Exception as e:
+                if "Target property type isn't nanobind type" in str(e):
```
njroussel, line 78:
```suggestion
                if "Target property type isn't a nanobind type" in str(e):
```

njroussel (approved):
Tiny typo, but then it's good to go :rocket: 

Very clean!

Issue #1253, opened 2024-08-06
`std::bad_cast` when passing invalid-shaped NumPy array to plugin


dvicini:
The following code produces `std::bad_cast` on the new master branch. This used to work in the previous version. Note that the shape of the NumPy array does not have the right shape here, so some form of error might be the right thing to do here. But maybe a more informative message could be thrown, pointing to the user to the source of error ("invalid shape")?
```py
import mitsuba as mi 
import numpy as np
mi.set_variant('llvm_ad_rgb')

mi.load_dict({'type': 'rgb', 'value': np.array([[1.0,0.5,0.3]])})
```

njroussel:
~Fixed in #1267.

PR #1267, opened 2024-08-13
xml_v: Improved reporting of failed Color3f initialization

rtabbara:
Proposed fix for #1253


njroussel (approved):
This is a decent compromise in my opinion. It at least allows users to understand which argument lead to a `bad_cast`.
 
Good to go.

PR #1273, opened 2024-08-15
Fix unique_ptr type for stream read

dvicini:
Minor change to switch to `std::unique_ptr<char[]>` in the stream bindings. The current code produces undefined behavior by using `std::unique_ptr<char> tmp(new char[size]);`. This causes a mismatch between new[] and delete (without []) operators


njroussel (approved):
Thanks, nice catch :+1:

PR #1275, opened 2024-08-16
Filesystem test: use tmp_path for writing files

dvicini:
The filesystem test uses the current path (`path_here`), which is the directory from which pytest is run. This is fine for testing path-related logic (e.g., `assert fs.is_directory(path_here)`), but not ideal when creating test files and directories.

This commit changes this test to only create new files and directories within a temporary folder (`tmp_path`). This decouples running this test from the permissions the user has on the current working directory.


njroussel (approved):
Note to future self: `tmp_path` is a built-in fixture in pytest

Thanks, the changes look good to me.

Issue #1126, opened 2024-04-04
How to sample sggx distribution in Python?


Vrroom:
## Summary

How do I sample from the SGGX distribution from Python. I'm trying to implement the SpongeCake BSDF, thus needing to sample micronormals from the SGGX distribution. 

I was trying: 

```
>>> mi.sggx_sample(sh, mi.Point2f([0.5, 0.5]), dr.llvm.ad.ArrayXf([[1, 1, 1, 0, 0, 0]]))
```

Mitsuba complains that:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: sggx_sample(): incompatible function arguments. The following argument types are supported:
    1. (sh_frame: mitsuba.llvm_ad_rgb.Frame3f, sample: mitsuba.llvm_ad_rgb.Point2f, s: drjit::Array<drjit::DiffArray<drjit::LLVMArray<float> >, 6ul>) -> mitsuba.llvm_ad_rgb.Normal3f
    2. (sh_frame: mitsuba.llvm_ad_rgb.Vector3f, sample: mitsuba.llvm_ad_rgb.Point2f, s: drjit::Array<drjit::DiffArray<drjit::LLVMArray<float> >, 6ul>) -> mitsuba.llvm_ad_rgb.Normal3f

Invoked with: Frame[
  s = [[1, -0, -0]],
  t = [[-0, 1, -0]],
  n = [[0, 0, 1]]
], [[0.5, 0.5]], [[1.0],
 [1.0],
 [1.0],
 [0.0],
 [0.0],
 [0.0]]
```

How do I construct this specific type: `drjit::Array<drjit::DiffArray<drjit::LLVMArray<float> >, 6ul>` in Python. 

## System configuration

```
System information:

  OS: Ubuntu 20.04.5 LTS
  CPU: Intel(R) Xeon(R) Platinum 8358 CPU @ 2.60GHz
  GPU: NVIDIA A10
  Python: 3.11.4 (main, Jul  5 2023, 13:45:01) [GCC 11.2.0]
  NVidia driver: 525.85.12
  CUDA: 11.8.89
  LLVM: 12.0.0

  Dr.Jit: 0.4.4
  Mitsuba: 3.5.0
     Is custom build? False
     Compiled with: GNU 10.2.1
     Variants:
        scalar_rgb
        scalar_spectral
        cuda_ad_rgb
        llvm_ad_rgb
```

## Description

See the summary

[...]

PR #1276, opened 2024-08-16
Fix Python bindings for SGGX distribution functions

rtabbara:
Workaround because `dr.Array6*` bindings don't exist


njroussel:
Replaces #1052

njroussel (approved):
Looks good, thanks for the extra tiny test :smile:

njroussel:
Fixes #1126

PR #1281, opened 2024-08-20
Fix stokes integrator for row-major matrices

Speierers:
This patch fixes the stokes integrator to account for the change of layout in Dr.Jit matrix type (column-major --> row-major).


rtabbara:
Thanks Sébastien!

Issue #1280, opened 2024-08-19
Properties or scene loading tries to convert 3x3 Tensor to Array3f


dvicini:
The following code prints a warning:

```
nanobind: implicit conversion from type 'TensorXf' to type 'drjit.scalar.Array3f64' failed!
```

```python
import mitsuba as mi
import numpy as np

mi.set_variant('llvm_ad_rgb')

class MyEmitter(mi.Emitter):

    def __init__(self, props):
        super().__init__(props)
        print(props['test'])

mi.register_emitter('myemitter', lambda props: MyEmitter(props))

emitter = mi.load_dict({'type': 'myemitter', 'test': mi.TensorXf(np.ones((3,3), dtype=np.float32))})
```

For some reason, it tries to convert the `mi.TensorXf` to `Array3f64`. I've tried using `mi.TensorXd` instead, but this didn't change anything either. I can't reproduce this when only using a `mi.Properties` object in isolation, so it seems at least somewhat related to scene loading?

I am not sure how problematic this warning is in practice, but this conversion attempt is a bit suspicious and hints at something not being 100% right here.

PR #1285, opened 2024-08-22
Disable nanobind implicit cast warnings during `load_dict`

rtabbara:
Motivated by #1280


njroussel (approved):
Looks good to me, fixes #1280 

thanks :rocket:

PR #1294, opened 2024-09-03
Ensure external threads always get a Logger and FileResolver

dvicini:
This PR ensures that any external thread that is registered with Mitsuba will have a valid logger and file resolver.

The Logger is re-used from the main thread, since it is thread safe. The file resolver is copied, as it is not thread safe, but light weight to copy.

This PR also adds two tests that exercise this newly added / fixed functionality.

### DIFF
--- a/src/core/python/logger.cpp
+++ b/src/core/python/logger.cpp
@@ -6,6 +6,12 @@
 
 /// Submit a log message to the Mitusba logging system and tag it with the Python caller
 static void PyLog(mitsuba::LogLevel level, const std::string &msg) {
+
+    // Do not log if no logger is available. This is consistent with the
+    // C++-side "Log" function in logger.h.
+    if (!Thread::thread()->logger())
+        return;
+
 #if PY_VERSION_HEX >= 0x03090000
     PyFrameObject *frame = PyThreadState_GetFrame(PyThreadState_Get());
     PyCodeObject *f_code = PyFrame_GetCode(frame);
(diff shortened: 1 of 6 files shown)


src/core/python/logger.cpp
```diff
@@ -29,18 +35,6 @@ static void PyLog(mitsuba::LogLevel level, const std::string &msg) {
     if (!name.empty() && name[0] != '<')
         fmt.insert(2, "()");
 
-    if (!Thread::thread()->logger()) {
```
merlinND, line 32: Maybe this could be replaced with an assert?

dvicini: I am not against that. My goal was to make it consistent with the C++ side here, which also just ignores this case:
https://github.com/mitsuba-renderer/mitsuba3/blob/master/include/mitsuba/core/logger.h#L140

merlinND: Understandable.. I guess it will crash  either way :D

dvicini: It should just ignore and not log anything at all, which is not too bad.

merlinND: Ahh sorry I didn't pay attention to the check you added above. Makes sense.

The case of `!Thread::thread()->logger()` should no longer happen by default now that you automatically bring over the main thread's logger, right?
If there's no logger, then it must be that the user set a thread environment that explicitly doesn't have a logger?

dvicini: Yes exactly. Not sure if that's something one would ever want?

In any case, there are still some other issues to be fixed with this PR. The current version does not shut down cleanly..

src/core/tests/test_thread.py
```diff
@@ -3,11 +3,41 @@
...
     thread.join()
+
+def test02_access_logger_from_new_thread(variant_scalar_rgb):
+    def print_to_log():
+        mi.set_log_level(mi.LogLevel.Info)
+        mi.Log(mi.LogLevel.Info, "Log from a thread environment.")
+
+    thread = Thread(target = print_to_log)
+    thread.start()
+    thread.join()
```
njroussel, line 22: Let's add an assert to make sure that the log was visible (you can captue it with the `capsys` fixture).

njroussel (approved):
One tiny request, the rest looks good to me!

I've checked all of the build logs, none of the environments have leaks when running the test suite :+1:

src/core/tests/test_thread.py
dvicini: Good idea. I now added a `mi.StreamAppender`. I could not get it to work with `capsys`, this one only would capture Python-side prints for me (seems to be a known limitation)

Issue #1149, opened 2024-04-26
Missed intersections (embree-related)


tomas16:
## Summary

I encountered a case where rays seem to pass through solid geometry. 2 of the rays are at fairly glancing angles, but 3 others aren't. In all 5 cases, the rays seem to hit the edge between 2 triangles.

## System configuration

System information:

  OS: 14.2
  CPU: b'Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz'
  Python: 3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:51:20) [Clang 16.0.6 ]
  LLVM: 16.0.6

  Dr.Jit: 0.4.4
  Mitsuba: 3.5.0
     Is custom build? False
     Compiled with: AppleClang 14.0.0.14000029
     Variants:
        scalar_rgb
        scalar_spectral
        llvm_ad_rgb

## Description

I've extracted the relevant geometry from the original mesh and created a standalone example. In the example, I compare to:
* `trimesh.ray.ray_pyembree.RayMeshIntersector`: trimesh uses Embree version 2. (I refer to this as `EmbreeIntersector`)
* `trimesh.ray.ray_triangle.RayMeshIntersector`: a python fallback, similar to mitsuba without Embree. (I refer to this as `PyIntersector`)

Coloring in the plots:
* red = mitsuba
* blue = trimesh EmbreeIntersector
* green = trimesh PyIntersector

![Screenshot 2024-04-25 at 4 45 37 PM](https://github.com/mitsuba-renderer/mitsuba3/assets/8531909/b1c70300-57a8-4490-b0f2-baefbfd4545a)

In above screenshot, the left and middle geometry are parts of a sphere. According to mitsuba, the rays hit on the inside. Comparing to trimesh, the reason is they first pass through a different triangle at a shallow angle.

On the right, there are 3 rays passing through a first surface and then hitting a surface behind it. trimesh correctly finds the intersections with the first surface.

Note that even the `EmbreeIntersector` in trimesh gets this right. But trimesh uses Embree 2, vs mitsuba's Embree 3.

Some close-ups:

![Screenshot 2024-04-25 at 3 34 43 PM](https://github.com/mitsuba-renderer/mitsuba3/assets/8531909/9aecd1e9-bdbb-4fa2-920a-5743ac4ee082)

[...]

PR #1300, opened 2024-09-11
Add support for Embree's robust intersection flag

dvicini:
Embree supports a fast and "robust" ray intersection mode. Mitsuba uses the default, fast mode. This is likely the best choice for general rendering tasks, but can cause issues in certain applications. 

For example, some people use Mitsuba purely to do some specialized ray tracing (for its speed & simplicity). In such cases, the ray tracing might not happen in in a "Monte Carlo setting", and rays may be spawned in a structured way. In that case, it can happen that rays perfectly hit edges between triangles, which is a known failure mode of Embree's fast mode (see also https://github.com/mitsuba-renderer/mitsuba3/issues/1149).

This PR adds support to set the robust mode at runtime using a boolean flag at the scene level. 

This also adds a test for the behavior. The test currently checks both failure and success of the ray tracing. I am not yet 100% sure if the failure will happen on all platforms (due to subtleties in the compilers & math eval) - to be seen once the CI runs.

I also haven't updated the documentation yet. I wasn't sure yet where would be the best place to describe scene level flags.

### DIFF
--- a/docs/src/plugin_reference.rst
+++ b/docs/src/plugin_reference.rst
@@ -107,3 +107,41 @@ These flags indicate whether the parameter is differentiable or not, or whether 
 discontinuities and thus needs special treatment.
 
 See :py:class:`mitsuba.ParamFlags` for their documentation.
+
+
+Scene-wide attributes
+---------------------
+
+The Scene object exposes scene-wide attributes. Currently, this functionality is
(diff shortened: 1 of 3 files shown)


wjakob:
This is great -- thank you, Delio!

PR #1305, opened 2024-09-16
Fix tonemap CLI

dvicini:
Fixes a minor issue with the tonemap CLI. The non-variant functions are no longer in the `scalar_rgb` module itself. Therefore, it's better to just set the scalar_rgb variant here explicitly.


wjakob:
Thanks!

PR #1307, opened 2024-09-17
Avoid always running stub generation

rtabbara:
First pass of stub generation is based on a specific variant so we have variant-specific types in the function signatures (e.g. `dr.llvm.Float`. As with Dr.Jit stub generation, we run a subsequent command to perform some regex manipulation of the stubs to change these types to `auto` (among a few other things).

Originally, the CMake script would delete the intermediary stubs after this regex manipulation. But the result is that the stub generation will be performed for every build which isn't ideal. This PR aims to avoid this.


njroussel (approved):
LGTM thanks

Issue #1288, opened 2024-08-23
Rendering Dielectric Materials with Non-Independent Sampling


Disgruntoad:
I'm fairly new to rendering so maybe this is a mistake on my part.

## Summary

All these renders are using the `cuda_spectral` variant.

I was attempting to render the banner 2 scene from https://mitsuba.readthedocs.io/en/stable/src/gallery.html

I replaced the material types for the background wall and the central object. The wall is a measured bsdf from EPFL's RGL lab (https://rgl.epfl.ch/materials) and the object is a simulated rough dielectric of glass interior & air exterior (ggx distribution).

```xml
  <!-- Materials -->                                                                                  
  <bsdf type="roughdielectric" id="mat-Material.005">                                                 
    <string name="int_ior" value="bk7"/>                                                              
    <string name="ext_ior" value="air"/>                                                              
    <string name="distribution" value="ggx"/>                                                         
  </bsdf>                                                                                             
                                                                                                      
  <bsdf type="measured" id="mat-Floor">                                                               
    <string name="filename" value="/data/raw/epfl_rgl_bsdf/colodur_azure_4e_spec.bsdf"/>              
  </bsdf>
```

I was running into high convergence samples per pixel for the independent sampler -- convergence occurs only after `spp=32768` which is a bit higher than I'd ideally set the value to. Other sampling methods should converge faster, so I attempted to render with them, but I'm getting some really odd artifacts. The distribution of artifacts varies slightly between samplers, but they all have the same odd characteristics. 

I include an example of the independent and stratified samplers, both generated with spp=1024, max_depth=256.

Independent, spp=1024:
![glass_air_independent_256depth_1024spp_1024spi](https://github.com/user-attachments/assets/f14e7f80-2be5-442a-a4a2-61cc3c37d577)

Stratified, spp=1024
![glass_air_stratified_256depth_1024spp_1024spi](https://github.com/user-attachments/assets/df64d1d0-2351-4d03-913c-baacbfd190b3)

Independent, spp=16384:
![glass_air_256depth_16384spp_128spi](https://github.com/user-attachments/assets/326ad2bd-cd52-4079-a3aa-9a27c5fa7d7e)

Stratified, spp=16384:
![glass_air_stratified_256depth_16384spp_256spi](https://github.com/user-attachments/assets/4a6a138b-9a70-4870-beb4-c27a76c7ceaf)

For reference, here is the stratified render with the object as a rough conductor of Cu (ggx distribution).

[...]

PR #1308, opened 2024-09-17
Fix passing sampler state within symbolic control flow

rtabbara:
Fix for #1288 . The reason why the independent sampler was fine was that we only had a single state variable so the interface for passing in the state wasn't affected.

Before and after comparison.

Stratified:
image

Ldsampler:
image

Multijitter:
image

Orthogonal:
image


njroussel (approved):
LGTM

I'm confused how any of these worked in the previous versions now. I can't figure out how we were passing all of these extra variables to the loop state...

PR #1317, opened 2024-09-24
CMake stubs: Change post-process command dependency to target-level 

rtabbara:
Motivated by flaky Windows CI, where we appear to be generating stubs twice, in some instances resulting in a race condition.

After stub generation through nanobind, we perform an additional custom command to regex variant specific types to auto. e.g. `dr.llvm.ad.Float` -> `dr.auto.ad.Float`. Change the cmake custom command dependency from the output `.pyi` file to the corresponding stub-gen target that gets created in `nanobind_add_stub`  to ensure correct execution ordering.


rtabbara:
Looking at the Windows CI build, it appears that indeed the Mitsuba `*.pyi` files are now only generated once

njroussel (approved):
Great!

PR #1319, opened 2024-09-24
Allow vectorized vcalls of Mesh methods

merlinND:
## Description

Allows accessing the `Mesh` interface through a new `MeshPtr` type. This will allow users to access the `Mesh` interface through a new `MeshPtr`. Before, it was only accessible on the individual `Mesh` instances (no vectorized calls).

```py
meshes = mi.MeshPtr(si.shape)
# Not strictly needed, the non-mesh pointers are automatically zeroed-out
active &= si.shape.is_mesh() 

n = meshes.face_normal(si.prim_index, active=active)
```

Requires a DrJit-side change to allow `CallSupport<Mesh>`: https://github.com/mitsuba-renderer/drjit/pull/287 

## Testing

Added a unit test that checks the results of the vcall on `MeshPtr` against calls on the individual mesh instances.

## Checklist

### DIFF
--- a/src/render/python/shape_v.cpp
+++ b/src/render/python/shape_v.cpp
@@ -268,6 +268,24 @@ template <typename Ptr, typename Cls> void bind_mesh_generic(Cls &cls) {
             },
             "index"_a, "ray"_a, "active"_a = true,
             D(Mesh, ray_intersect_triangle));
+
+    if constexpr (dr::is_array_v<Ptr> && dr::is_jit_v<Ptr>) {
+        // Custom constructor to automatically zero-out non-Mesh pointer entries.
+        // using ShapePtr = dr::replace_scalar_t<Ptr, Shape *>;
+        cls
+            .def("__init__", [](Ptr *dst, const ShapePtr &ptr) {
+                ShapePtr filtered = dr::select(ptr->is_mesh(), ptr, dr::zeros<ShapePtr>());
+                Ptr mesh = dr::reinterpret_array<Ptr>(filtered);
+                // Placement new.
+                new (dst) Ptr(mesh);
+            })
+            .def("__init__", [](Ptr *dst, const Shape *ptr) {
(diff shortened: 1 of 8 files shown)


src/render/python/shape_v.cpp
```diff
@@ -217,6 +219,75 @@ template <typename Ptr, typename Cls> void bind_shape_generic(Cls &cls) {
...
+                ptr->ray_intersect_triangle(index, ray, active);
+            },
+            "index"_a, "ray"_a, "active"_a = true,
+            D(Mesh, ray_intersect_triangle));
+
+    if constexpr (dr::is_array_v<Ptr> && dr::is_jit_v<Ptr>) {
+        // Custom constructor to automatically zero-out non-Mesh pointer entries.
+        // using ShapePtr = dr::replace_scalar_t<Ptr, Shape *>;
+        cls
+            .def("__init__", [](Ptr *dst, const ShapePtr &ptr) {
```
merlinND, line 276: Ideally I would have wanted to add extra constructors for `const ShapePtr &` and leave the basic ones as-is, but I don't think it's possible?

njroussel: I'm not sure I understand. Are you talking about `.def("__init__", [](Ptr *dst, const Mesh *ptr)` ? As in, it should already exist?

Although unfortunate, this is expected because of how the original Dr.Jit contstructors are defined (they are defined with [type slots](https://nanobind.readthedocs.io/en/latest/typeslots.html)). They will be lost when using the `def("__init__", ... )` mechanism.

merlinND: Okay, makes sense! It's not too much code to add back the constructors we need.

merlinND:
Thank you! I rebased and applied the fix.

njroussel:
Looks good now. Thank you!

PR #1326, opened 2024-10-01
Change seed function to take drjit type for seed

DoeringChristian:
## Description

This PR changes the signature of the `seed` function for samplers (and their implementations) to take a Dr.Jit `UInt32` instead of a `uint32_t` as the seed.
It also changes the signature `compute_per_sequence_seed` function to take the same type as the argument.
The change will eventually be necessary for function freezing, as it allows passing already opaque seeds to the render function and thereby prevents "re-recording" of the frozen function.

## Checklist

- [ ] I give permission that the Mitsuba 3 project may redistribute my contributions under the terms of its [license](https://github.com/mitsuba-renderer/mitsuba/blob/master/LICENSE)

### DIFF
--- a/include/mitsuba/render/sampler.h
+++ b/include/mitsuba/render/sampler.h
@@ -95,7 +95,7 @@ public:
      * In the context of wavefront ray tracing & dynamic arrays, this function
      * must be called with \c wavefront_size matching the size of the wavefront.
      */
-    virtual void seed(uint32_t seed,
+    virtual void seed(UInt32 seed,
                       uint32_t wavefront_size = (uint32_t) -1);
 
     /**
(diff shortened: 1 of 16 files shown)


njroussel (approved):
LGTM !

@DoeringChristian  thanks. One quick question before I merge: I assume we don't need to change `m_base_seed` because it won't be part of a function/interface that can be freezed ?

DoeringChristian:
> LGTM !
> 
> @DoeringChristian thanks. One quick question before I merge: I assume we don't need to change `m_base_seed` because it won't be part of a function/interface that can be freezed ?



> LGTM !
> 
> @DoeringChristian thanks. One quick question before I merge: I assume we don't need to change `m_base_seed` because it won't be part of a function/interface that can be freezed ?

Good point. I think its not necessary for now. As it is now, the `base_seed` will be frozen and calling the frozen function with a sampler with a different `base_seed` would use the old one. We have a few of these cases, where C++ types are part of some class that is an argument to a frozen function and a change to this value will be disregarded by the freezing alrorithm. I think we decided to ignore those cases for now.

njroussel:
Ok, thanks for the clarification!

PR #1344, opened 2024-10-15
Bump Dr.Jit version

wjakob:
Check to see if the Mitsuba test suite passes following some Dr.Jit internals changes related to horizontal reductions.


njroussel:
Already in the `master` branch.

PR #1329, opened 2024-10-04
Updating Python wheels workflow for Mitsuba 3.6.0 release

rtabbara:
- Use native arm64 Github runner and transfer CIBW envvars to `pyproject.toml`
- Build for also Python 3.13, stable ABI Python 3.12
- Make sure we install generated stubs
- Gracefully handle static initialization of Mitsuba for stub generation when LLVM/CUDA is missing

Note: Dr.Jit submodule change is temporary to test wheels and will be updated once [Dr.Jit PR 293](https://github.com/mitsuba-renderer/drjit/pull/293) is merged

PR #1355, opened 2024-10-23
Updated PyPI release process

njroussel:
This PR supersedes https://github.com/mitsuba-renderer/mitsuba3/pull/1329, which originally made the following changes

>* Use native arm64 Github runner and transfer CIBW envvars to pyproject.toml
 > * Build for also Python 3.13, stable ABI Python 3.12
  >* Make sure we install generated stubs
  >* Gracefully handle static initialization of Mitsuba for stub generation when LLVM/CUDA is missing

in this PR we additionally have the following:

* Nanobind is built as a shared library
* Wheel repair (on Linux) and testing is simplified
* The set of variants built for PyPI is updated (see below)
* Minor configuration changes for depedency requirements

The new list of variants that will  be available on PyPI:
* scalar_rgb
* scalar_spectral
* scalar_spectral_polarized
* llvm_ad_rgb
* llvm_ad_mono
* llvm_ad_mono_polarized
* llvm_ad_spectral
* llvm_ad_spectral_polarized
* cuda_ad_rgb
* cuda_ad_mono
* cuda_ad_mono_polarized
* cuda_ad_spectral
* cuda_ad_spectral_polarized


rtabbara (approved):
LGTM :rocket: Don't think it makes sense in this PR, but we need to update the `README.md` to reflect the change to the variants included with the Python package just before we generate the wheels.

PR #1362, opened 2024-10-25
Use variant-specific registry domains

merlinND:
## Description

Previously, we could easily segfault when performing a vcall if an object was previously created with another variant. For example, a `spectral_polarized` BSDF and an `rgb` BSDF.

Other than fixing the case below, I don't think this change could hurt since we never need or want instances from different variants to participate in the same vcalls (?).

## Testing

```py
import drjit as dr
import mitsuba as mi

def test01_test_variants_switching():
    mi.set_variant('cuda_ad_mono_polarized')
    # Unused, but registers objects into the registry
    scene1 = mi.load_dict(mi.cornell_box())

    mi.set_variant('cuda_ad_rgb')
    scene2 = mi.load_dict(mi.cornell_box())
   
    # Before this PR, this would segfault when e.g. collecting the results of the BSDF vcall,
    # because the `Spectrum` types of the BSDF instances were not compatible.
    image = mi.render(scene2)
    print(image)


if __name__ == "__main__":
    test01_test_variants_switching()
```

## Checklist

### DIFF
--- a/include/mitsuba/core/class.h
+++ b/include/mitsuba/core/class.h
@@ -291,8 +291,10 @@ constexpr auto domain_name_for_class(const std::array<char, N> &class_name) {
     DRJIT_CALL_TEMPLATE_BEGIN(mitsuba::Name)
 
 #define MI_CALL_TEMPLATE_END(Name)                                                      \
+private:                                                                                \
     static constexpr std::array<char, sizeof(#Name) / sizeof(char)> MIClass_{#Name};    \
     static constexpr auto MIDomain_ = mitsuba::domain_name_for_class<Ts...>(MIClass_);  \
+public:                                                                                 \
     static constexpr const char *domain_() {                                            \
         return MIDomain_.data();                                                        \
     }                                                                                   \
--- a/include/mitsuba/core/class.h
+++ b/include/mitsuba/core/class.h
@@ -280,22 +280,18 @@ constexpr auto domain_name_for_class(const String &class_name) {
 /**
  * Adds the given class instance to a DrJit registry domain that is specific
  * to this backend and Mitsuba variant.
+ *
+ * Example usage:
+ *     MI_REGISTRY_PUT("BSDF", this);
  */
-#define MTS_REGISTRY_PUT(name, ptr)                                                       \
-    if constexpr (dr::is_jit_v<Float>) {                                                  \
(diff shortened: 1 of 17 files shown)


merlinND:
This ended up being a fair bit more complicated that I expected, because now that the domain name depends on the `Float, Spectrum` template arguments, it cannot directly be set in the call support macros from the `Name` macro parameter.

But because `call_support<>::Domain` is used as a static constexpr field in a few places, we still have to compute the new name in a constexpr way.
In C++17, it's possible to do constexpr string concatenation via `std::array<>`.

If we find a simpler solution (either to the original issue or any of the implementation steps of this PR), I would be all for it.

njroussel (changes requested):
I agree that we should have some kind of a safety net for this issue.

Your suggested implementation seems fine. Changing the `Domain` is what makes the most sense, in my opinion.

Could you change the macros to be prefixed with `MI_` instead of `MTS_`. 

Also the CI is currently failing due to the `variant_str` declaration being non-constexpr. I believe that you should be able to fix that with a `std::array` again.

merlinND:
Thanks for having a look @njroussel!

I updated the macros to use `MI_` instead of `MTS` and used `domain_()` instead of `__domain()` (matching the other DrJit methods like `abs_`, `sin_`, etc).

I tried another implementation of the constexpr string concatenation (using more `std::array<char, N>`) that will hopefully work across compilers.

merlinND:
CI fails look like:
- MVSC: internal compiler error (`  C:\BuildAgent\work\c1e5bf5b71462fab\include\mitsuba\core\class.h(254): fatal error C1001: Internal compiler error. [C:\BuildAgent\work\c1e5bf5b71462fab\build\src\mitsuba.vcxproj]`)
- Others: `volpath` rendering tests failing. I don't know what's special about volpath since the other rendering tests are passing. I checked `Medium` and `PhaseFunction`, and they are the same as the other interfaces w.r.t. this new variant-specific domain. Maybe `volpath` is the only class in Mitsuba that triggers recursive vcalls, and that now somehow fails? :thinking:

merlinND:
Hi @njroussel,

Does the current state of this PR (+ the smaller one on the DrJit side: https://github.com/mitsuba-renderer/drjit/pull/307) look good to you?

njroussel (approved):
Thanks for the changes! 
I've added a naive test - based on what you had put in the description of the PR.

Thank you!

PR #1367, opened 2024-10-28
 Variants: allow registering callbacks for variant changes

merlinND:
## Description

As discussed with @njroussel.

Currently, upon each variant change, we explicitly reload the custom Python integrators shipped in `mitsuba.python.ad.integrators`.
Users don't have access to such a mechanism.

By allowing users to register callbacks, they can now implement module loading & reloading for their custom Python plugins.
In the future, there may be a higher-level way for users to register plugins that will automatically create such a reload callback.

I also added an alias `scoped_set_variant` to `variant_context` so that it matches `dr.scoped_set_flag()`.

## Testing

Added a unit test check.

## Checklist


src/python/alias.cpp
```diff
@@ -112,23 +116,56 @@ static void set_variant(nb::args args) {
...
             if (!nb::bool_(k.attr("startswith")("__")) &&
                 !nb::bool_(k.attr("endswith")("__")))
                 Safe_PyDict_SetItem(mi_dict, k.ptr(),
                     PyDict_GetItem(variant_dict.ptr(), k.ptr()));
 
+        const auto &callbacks = nb::borrow<nb::set>(variant_change_callbacks);
+        for (const auto &cb : callbacks)
+            cb(curr_variant, new_variant);
+
+        // TODO: replace this with a callback?
```
merlinND, line 132: We could use the opportunity to remove this special case and add a callback in `mitsuba.python.ad.integrators.__init__`. But I wasn't sure how to test the new behavior is correct.

Does the test suite check / relies on this reload to correctly happen?

src/python/alias.cpp
```diff
@@ -178,17 +210,24 @@ NB_MODULE(mitsuba_alias, m) {
     /// Only used for variant-specific attributes e.g. mi.scalar_rgb.T
     m.def("__getattr__", [](nb::handle key) { return get_attr(key); });
 
+    // `mitsuba.detail` submodule
+    nb::module_ mi_detail = m.def_submodule("detail");
```
merlinND, line 214: @wjakob At first I tried to `nb::module_::import_("mitsuba.detail");` but there was no such module, please let me know if this is not the proper way.

wjakob: That looks good to me!

PR #1383, opened 2024-11-08
AOV Integrator: Fix AD computation

rtabbara:
By default, Dr.Jit’s AD system destructs the enqueued input graph during AD traversal. But the AOV integrator `render_forward/render_backward` functions consists of separate `dr::forward_to`/`dr::backward_from` calls for the AOV channels and inner integrator output. Hence we need to perform repeated propagation of gradients of a shared AD subgraph.


njroussel (approved):
:100:

PR #1415, opened 2024-11-27
Fix mi.sample_tea_float return type

tomix1024:
## Description

When calling `mi.sample_tea_float` the delegated function was chosen incorrectly, always selecting the `Float64` version.
On the C++ side the opposite happened, always selecting the `Float32` version.

The python fix is straight forward.
On the C++ side I added the result `Float` type as an additional template argument, because I did not see any other way to query the equivalent of `mi.Float` reliably. `MI_VARIANT_FLOAT` is not always to be defined.

## Testing

I have added `test08_sample_tea_float` to `test_random.py` to test the return type in the Python interface.

## Checklist


rtabbara:
LGTM! :rocket: Thanks @tomix1024 !

Issue #1051, opened 2024-01-25
Bug in Adam for non-real parameters


dvicini:
Hi,

There appears to be an edge case in the current Adam implementation. If the optimizer contains non-real variables, i.e., quaternions or complex numbers, the variance estimation might not be what we want.

The Adam implementation computes
```
v_t = self.beta_2 * v_tp + (1 - self.beta_2) * dr.sqr(g_p)
```

where `g_p` is the parameter gradient. If a parameter is a quaternion or complex number, the gradient will be of the same type. Then, `dr.sqr` will compute a quaternion or complex number product. This is different from the element-wise product that we would expect here.

While there is some literature on this subject ([pytorch github issue](https://github.com/pytorch/pytorch/issues/59998), [pytorch docs](https://pytorch.org/docs/stable/notes/autograd.html#autograd-for-complex-numbers), https://arxiv.org/pdf/0906.4835.pdf), the easiest practical solution is to just optimize using Vector2f or Vector4f instead.

Maybe Mitsuba should raise an error if a quaternion or complex value is assigned to the optimizer?

njroussel:
Hi @dvicini 

I agree with you. My suggestion would be a check to see if the type is special (`dr.is_special_v()`) and  if it is the case raise an error as you said.

I'll leave this open for a few more days before pushing a fix, in case anyone else has some ideas to contribute :)

wjakob:
The same issue would also appear with camera matrices and, e.g., complex IOR values in the next version.

The optimizer will need an extra check like

```cpp
tp = type(g_p)
if dr.is_special_v(tp):
    v_t = self.beta_2 * v_tp + (1 - self.beta_2) * dr.sqr(g_p)
else:
    tp = dr.array_t(tp)
    v_t = self.beta_2 * v_tp + (1 - self.beta_2) * dr.sqr(tp(g_p))
```

wjakob:
We could either wait until the next version or backport the `dr.array_t` type trait.

rtabbara:
This has now been fixed in #1051

PR #1423, opened 2024-12-02
Fix for #1051 Adam for non-real parameters

rtabbara:
Proposed fix as mentioned in the issue was reliant on Mitsuba 3.6.0 release (Dr.Jit 1.0.0). We can now use `dr.array_t`


src/python/python/ad/optimizers.py
```diff
@@ -279,6 +279,13 @@ def step(self):
 
             m_tp, v_tp = self.state[k]
             m_t = self.beta_1 * m_tp + (1 - self.beta_1) * g_p
+
+            # When computing variance, want to take element-wise product of g_p
```
dvicini, line 283: Maybe write:

"Ensure that we take the element-wise product to compute variance"

PR #1452, opened 2025-01-10
Remove unnecessary suspend_grad

dvicini:
PRB suspends gradients on sample positions, ray and ray weight in the backward pass. I think this might be a leftover from the time we supported reparameterized PRB, where it was important to carefully detach the splatting position. Since the basic PRB integrator generally does not support differentiating sensor pose, I don't think this is needed anymore. Projective sampling anyway uses a separate base integrator class.

We came across this when trying to optimize a scale on the ray weight (e.g., an exposure parameter), which did not work due to this suspend_grad context.


njroussel (approved):
Yes indeed, this was only meant for a moving sensor.

PR #1477, opened 2025-01-27
Frozen Function Support

DoeringChristian:
## Description

This PR adds `TRAVERSE_CB` macros to most classes in Mitsuba. This allows the frozen function feature to traverse a scene and enables frozen function support (see [here](https://github.com/mitsuba-renderer/drjit/pull/317)).
Additionally, it expands on #1326 to pass UInt32 type seeds, allowing rendering scenes in frozen functions with different seeds without re-tracing.

The following is a more concise list of changes:
- `Object` is now inheriting from `drjit::TraversableBase` and implements the `traverse_1_cb_ro` and `traverse_1_cb_rw` functions.
- Classes that can be added to the registry now use `drjit::registry_put` instead of `jit_registry_put`, enforcing that they inherit from `drjit::TraversableBase` at compile time.
- Most classes now implement `traverse_1_cb_ro` and `traverse_cb_rw` from the `drjit::TraversableBase` class using either a combination of the macros `DR_TRAVERSE_CB`, `MI_DECLARE_TRAVERSE_CB`, and `MI_IMPLEMENT_TRAVERSE_CB`.
- Implements the latter two macros in Mitsuba.
- Makes the optimizer classes traversable by adding a `DRJIT_STRUCT` property
- Adds special traversal functions for `Scene`s and it's subtypes (GPU and CPU)

## Testing

Added tests for shapes, bsdfs, emitters, integrators, optimizers, meidia and samplers as well as tests verifying gradient descent in frozen functions (pose and material parameter optimization).

## Checklist


src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+
+        return image_final, opt["trans"], opt["angle"]
+
+    n = 10
+
+    # NOTE:
+    # In this case, we have to use the same scene object
+    # for the frozen and non-frozen case, as re-loading
+    # the scene causes mitsuba to render different images,
+    # leading to diverging descent traijectories.
```
merlinND, line 241:
```suggestion
    # leading to diverging descent trajectories.
```

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+        assert dr.allclose(img_ref, img_frozen, atol=1e-4)
+
+
+def test03_optimize_color(variants_vec_rgb):
+    k = "red.reflectance.value"
+    w = 16
+    h = 16
+    n = 10
+
+    def mse(image, image_ref):
```
merlinND, line 270: Factor out

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+                "type": bsdf,
+            }
+        scene = mi.load_dict(scene, parallel=False)
+        return scene
+
+    scene = load_scene(bsdf)
+    images_ref = run(scene, n, func)
+    scene = load_scene(bsdf)
+    images_frozen = run(scene, n, frozen)
+
```
merlinND, line 570: Maybe add a couple of direct function calls on the BSDFs (including the custom one). Rendering uses the BSDFs through vcalls, but it would be good to check direct calls of `eval()`, `pdf()`, `sample()`.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+            images.append(img)
+
+        return images
+
+    images_ref = run(n, func)
+    # scene = load_scene(emitter)
+    # del scene
+    # gc.collect()
+    # gc.collect()
```
merlinND, line 701:
```suggestion
```

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+    images_frozen = run(n, frozen)
+
```
merlinND, line 703: Same as above, testing direct calls.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+        "projector",
+        "directional",
+        "directionalarea",
+    ],
+)
+def test05_emitter(variants_vec_rgb, emitter):
+    # dr.set_log_level(dr.LogLevel.Trace)
+    # dr.set_flag(dr.JitFlag.ReuseIndices, False)
+    # dr.set_flag(dr.JitFlag.Debug, True)
+
```
merlinND, line 593:
```suggestion
```

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+                "type": "prb_basic",
+            }
+        elif integrator == "direct_projective":
+            scene["integrator"] = {
+                "type": "direct_projective",
+            }
+        elif integrator == "prb_projective":
+            scene["integrator"] = {
+                "type": "prb_projective",
+            }
```
merlinND, line 769:
```suggestion
        scene["integrator"] = { "type": integrator }
```

and then handle only the special cases in `if / else` blocks.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+@pytest.mark.parametrize(
+    "shape",
+    [
+        "mesh",
+        "disk",
+        "cylinder",
+        "bsplinecurve",
+        "linearcurve",
+        "sdfgrid",
+        # "instance",
```
merlinND, line 818: Are `instance` shapes not supported?
If we don't manage to supported them, we should find a way to throw an exception if the user attempts to use them.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+
+        return images
+
+    scene = load_scene()
+    images_ref = run(scene, n, func)
+    images_frozen = run(scene, n, frozen)
+
+    assert frozen.n_recordings < n
+    for i, (ref, frozen) in enumerate(zip(images_ref, images_frozen)):
+        os.makedirs(f"out/{shape}", exist_ok=True)
```
merlinND, line 959: Use e.g. `tempfile.gettempdir()` or an equivalent feature in `pytest`, and don't write out the image files by default.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+        mi.util.write_bitmap(f"out/{shape}/ref{i}.jpg", ref)
+        mi.util.write_bitmap(f"out/{shape}/frozen{i}.jpg", frozen)
```
merlinND, line 961: Save as EXR

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+    # scene = load_scene()
+    images_ref = run(n, func)
+    images_frozen = run(n, frozen)
+
+    assert frozen.n_recordings < n
+
+    for i, (ref, frozen) in enumerate(zip(images_ref, images_frozen)):
+        os.makedirs(f"out/{medium}", exist_ok=True)
+        mi.util.write_bitmap(f"out/{medium}/ref{i}.jpg", ref)
+        mi.util.write_bitmap(f"out/{medium}/frozen{i}.jpg", frozen)
```
merlinND, line 1103: Same comments as above

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+
+    assert frozen.n_recordings < n
+
+    for i, (ref, frozen) in enumerate(zip(images_ref, images_frozen)):
+        os.makedirs(f"out/{sampler}", exist_ok=True)
+        mi.util.write_bitmap(f"out/{sampler}/ref{i}.jpg", ref)
+        mi.util.write_bitmap(f"out/{sampler}/frozen{i}.jpg", frozen)
+
+    for ref, frozen in zip(images_ref, images_frozen):
+        assert dr.allclose(ref, frozen)
```
merlinND, line 1170: Missing newline at end of file (you can configure your editor so that it always adds one).

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+    # scene = load_scene()
+    images_ref = run(n, func)
+    images_frozen = run(n, frozen)
+
+    assert frozen.n_recordings < n
+
+    for i, (ref, frozen) in enumerate(zip(images_ref, images_frozen)):
+        os.makedirs(f"out/{sampler}", exist_ok=True)
+        mi.util.write_bitmap(f"out/{sampler}/ref{i}.jpg", ref)
+        mi.util.write_bitmap(f"out/{sampler}/frozen{i}.jpg", frozen)
```
merlinND, line 1167: Same comment as above.

src/render/tests/test_freeze.py
merlinND: Same below for other kinds of objects (except maybe integrators, which are already tested directly by `render()`).

merlinND (commented):
Very nice PR! I like that the tests systematically cover a lot of the plugins.
Given the number of new tests, we may have to check that they don't slow down the CI too much.

Mostly small comments & consistency fixes, the bigger points are about:
- How to name the traverse macros
- Ways to improve the tests a bit

src/python/python/ad/integrators/prb.py
```diff
@@ -98,9 +98,13 @@ def sample(self,
...
 
-        while dr.hint(active,
-                      max_iterations=self.max_depth,
-                      label="Path Replay Backpropagation (%s)" % mode.name):
+        while dr.hint(
+            active,
+            max_iterations=self.max_depth,
+            label="Path Replay Backpropagation (%s)" % mode.name,
+            # include=[ sampler, ray, depth, L, δL, β, active ],
+            exclude=[ scene, ],
```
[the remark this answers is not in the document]
DoeringChristian: With the addition of the freezing traversal flag, should I remove this instead?

src/render/tests/test_freeze.py
DoeringChristian: Sorry, I'm not quite sure what you mean by factor out in this case? Should I put it in global scope?

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+            }
+        elif integrator == "ptracer":
+            scene["integrator"] = {
+                "type": "ptracer",
+                "max_depth": 8,
+            }
+        elif integrator == "depth":
+            scene["integrator"] = {"type": "depth"}
+
+        scene = mi.load_dict(scene, parallel=False)
```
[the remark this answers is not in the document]
DoeringChristian: That works as well, I just added this for debugging purposes. Should I add a global toggle?

include/mitsuba/core/fwd.h
```diff
@@ -368,6 +368,30 @@ extern "C" {
                             })
 #endif
 
+#define MI_DECLARE_TRAVERSE_CB()                                           \
```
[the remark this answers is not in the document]
DoeringChristian: Found a way to do this. If `MI_DECLARE_TRAVERSE_CB` defines a `traverse_1_cb_fields_` function (similar to `DRJIT_STRUCT`) we can traverse this in the implementation. Now we can list the members in the `DECLARE` macro and no longer have to list them in the `IMPLEMENT` macro.

src/shapes/bsplinecurve.cpp
```diff
@@ -1310,6 +1310,8 @@ class BSplineCurve final : public Shape<Float, Spectrum> {
     mutable CUdeviceptr* m_radius_buffer_ptr = nullptr;
     mutable CUdeviceptr* m_index_buffer_ptr = nullptr;
```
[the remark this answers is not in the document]
DoeringChristian: This should work, as they are just pointers referring to points in the `m_control_points` and `m_indices` buffers, which are traversed. I added a test to update the bspline every iteration.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+            images.append(img)
+
+        return images
+
+    scene = load_scene()
+    images_ref = run(scene, n, func)
+    scene2 = load_scene()
+    images_frozen = run(scene2, n, frozen)
+
+    assert frozen.n_recordings < n
```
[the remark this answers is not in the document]
DoeringChristian: I'm checking for `n_recordings == 2` now. Two recordings are produced, since the first iteration produces a different signature, compared to further iterations.

src/render/tests/test_freeze.py
```diff
@@ -0,0 +1,1170 @@
...
+            images.append(img)
+
+        return images
+
+    images_ref = run(n, func)
+    images_frozen = run(n, frozen)
+
+    assert frozen.n_recordings < n
+    for ref, frozen in zip(images_ref, images_frozen):
+        assert dr.allclose(ref, frozen)
```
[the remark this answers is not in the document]
DoeringChristian: Factored this out for most tests except optimization based ones.

src/render/tests/test_freeze.py
DoeringChristian: Can I also remove them?

include/mitsuba/core/fwd.h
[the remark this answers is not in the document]
DoeringChristian: Since the members are now listed in the `MI_DECLARE_TRAVERSE_CB` macro, we would have to add them to the `MI_DECLARE_CLASS` macro. This might be a bit confusing.

src/integrators/moment.cpp
```diff
@@ -141,6 +141,8 @@ class MomentIntegrator final : public SamplingIntegrator<Float, Spectrum> {
 private:
     std::vector<std::string> m_aov_names;
     std::vector<std::pair<ref<Base>, size_t>> m_integrators;
+
+    DR_TRAVERSE_CB(Base, m_integrators);
```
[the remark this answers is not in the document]
DoeringChristian: Added a test to traverse this type in `test_custom_type_ext.py` in `drjit`.

src/python/python/util.py
```diff
@@ -354,14 +354,16 @@ def eval(self, scene, sensor, _, params, integrator, seed, spp):
...
-            return self.integrator.render(
+            res = self.integrator.render(
                 scene=self.scene,
                 sensor=sensor,
                 seed=seed[0],
                 spp=spp[0],
                 develop=True,
                 evaluate=False
             )
+            sensor.sampler().seed(0, 1)
```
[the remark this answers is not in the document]
DoeringChristian: Should we possibly make this conditional on if we are in a freezing scope? It should not make a difference, but might indicate the reason a bit better.
```python
            if dr.flag(dr.JitFlag.FreezingScope):
                sensor.sampler().seed(0, 1)
```

src/render/tests/test_freeze.py
DoeringChristian: Thanks for pointing this out! Discovered, that I was not traversing cylinders correctly, because I implemented a test for `sample_position` for shapes.

merlinND:
Hi @DoeringChristian,

FYI, I think the PR was auto-closed when @njroussel deleted the `ser` branch after merging its contents to `master`. I temporarily re-created an `ser` branch so that I could re-open this PR and change its base back to `master`.
I think it'll just need one more rebase to `master` now.

PR #1478, opened 2025-01-27
Re-enable FP16 support for bitmap textures

rtabbara:
* Prior to Mitsuba 3.6, FP16 storage for bitmaps was disabled due to the observed degradation in performance (particularly on LLVM backends) relative to the single precision counterpart
* Changes with Dr.Jit textures now using PacketOps (https://github.com/mitsuba-renderer/drjit/pull/329) means we can re-enable support for FP16 bitmap textures


njroussel (approved):
LGTM :+1: 

I've just updated the sumbodule to point to `drjit:master`

PR #1484, opened 2025-02-05
Shape: add `set_bsdf()` method

merlinND:
## Description

This PR adds a simple `set_bsdf()` method to the `Shape` interface, which allows users to easily modify BSDFs after scene loading.
Note that this approach is restricted to BSDFs and was added after alternatives were investigated thoroughly. Deeper modifications to the scene graph after it is loaded is not supported.

## Testing

Added a unit test.

## Checklist

### DIFF
--- a/include/mitsuba/render/mesh.h
+++ b/include/mitsuba/render/mesh.h
@@ -17,8 +17,8 @@ template <typename Float, typename Spectrum>
 class MI_EXPORT_LIB Mesh : public Shape<Float, Spectrum> {
 public:
     MI_IMPORT_TYPES(BSDF)
-    MI_IMPORT_BASE(Shape, set_bsdf, m_to_world, mark_dirty, m_emitter, m_sensor,
-                   m_bsdf, m_interior_medium, m_exterior_medium, m_is_instance,
+    MI_IMPORT_BASE(Shape, m_to_world, mark_dirty, m_emitter, m_sensor, m_bsdf,
+                   m_interior_medium, m_exterior_medium, m_is_instance,
                    m_discontinuity_types, m_shape_type, m_initialized)
 
     // Mesh is always stored in single precision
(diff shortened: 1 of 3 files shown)


include/mitsuba/render/mesh.h
```diff
@@ -16,9 +16,9 @@ NAMESPACE_BEGIN(mitsuba)
 template <typename Float, typename Spectrum>
 class MI_EXPORT_LIB Mesh : public Shape<Float, Spectrum> {
 public:
-    MI_IMPORT_TYPES()
-    MI_IMPORT_BASE(Shape, m_to_world, mark_dirty, m_emitter, m_sensor, m_bsdf,
-                   m_interior_medium, m_exterior_medium, m_is_instance,
+    MI_IMPORT_TYPES(BSDF)
+    MI_IMPORT_BASE(Shape, set_bsdf, m_to_world, mark_dirty, m_emitter, m_sensor,
```
wjakob, line 20: Does it make sense to import `set_bsdf` if we override it below?

src/render/python/shape_v.cpp
```diff
@@ -301,6 +301,7 @@ MI_PY_EXPORT(Shape) {
         .def_method(Shape, id)
         .def_method(Shape, is_mesh)
         .def_method(Shape, parameters_grad_enabled)
+        .def_method(Shape, set_bsdf, "bsdf"_a.none())
```
wjakob, line 304: Do we support `nullptr` as BSDF? I suspect that this would crash the rendering process, at least in scalar mode. By default, a diffuse BRDF is created when none is specified.

wjakob (commented):
This looks good, just some minor comments.

src/render/python/shape_v.cpp
merlinND: Good point. I added this thinking "might as well", but I didn't have a precise use-case in mind. I'll remove it for safety.

Speierers:
Love it, thanks @merlinND!

njroussel (approved):

PR #1495, opened 2025-02-18
[CI run] No longer require libc++ when using Clang

merlinND:
## Description

```source ./.venv/bin/activate
pip install matplotlib==3.10.0

# Mitsuba built from `master` with Clang (--> libc++)
source ./build-Release/setpath.sh

python -c "import mitsuba as mi; from matplotlib import ft2font"
 [1]    40526 segmentation fault (core dumped)
```

When importing the module, two attributes that don't exist are looked up: `__path__` and `__cached__`. These [are optional](https://docs.python.org/3.12/reference/datamodel.html#module.__cached__), so it's fine that they don't exist.
But the way that `ft2font__getattr__` is implemented in the latest version leads to an exception being raised when they get looked up.
If we compile Mitsuba with GCC, libstdc++ is used, so the exception is correctly caught and handled and all is fine.
But with libc++, the exception thrown by ft2font is ABI-incompatible with the libc++ exception handlers, which results in a crash.

This is another instance of the classic `libc++` / `libstdc++` ABI incompatibility that we have run into multiple times before.
After some research, I wasn't able to find a way to fully isolate the exception handling routines of `libunwind` from exceptions thrown by other libraries built with `libstdc++`. After discussion with @wjakob, the simplest solution is then to stop requiring `libc++` to be used in conjunction with Clang (outside of macOS, where this is the default).

## Testing

CI run.

## Checklist

### DIFF
--- a/docs/src/developer_guide/compiling.rst
+++ b/docs/src/developer_guide/compiling.rst
@@ -98,7 +98,7 @@ TLDR: If you plan to use Mitsuba from Python, we recommend adding one of
 
     Note that compilation time and compilation memory usage is roughly
     proportional to the number of enabled variants, hence including many of them
-    (more than five) may not be advisable. Also note that the ``scalar_rgb`` 
+    (more than five) may not be advisable. Also note that the ``scalar_rgb``
     and *at least one AD variant* is mandatory.
 
 .. warning::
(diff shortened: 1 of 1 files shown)


njroussel (approved):
Thank you @merlinND !

PR #1522, opened 2025-03-18
set_variant(): try next requested variant if one fails to import

merlinND:
## Description

If the user specifies several variants in order, e.g.:

```py
mi.set_variant("cuda_ad_rgb", "llvm_ad_rgb")
```

and they are all compiled (available) but some fail to import, then keep trying with the next requested variant instead of throwing an exception.

This could happen e.g. when requested a CUDA variant with the NVIDIA driver installed / CUDA available but no GPU available, or when LLVM is not installed.

## Testing

On a machine with the NVIDIA driver installed,

```bash
CUDA_VISIBLE_DEVICES= python -c 'import mitsuba as mi; mi.set_variant("cuda_ad_rgb", "llvm_ad_rgb"); print(mi.variant())'
```

Old behavior: raises `ImportError` and exits.
New behavior: fails to load `cuda_ad_rgb`, but correctly goes to the second choice `llvm_ad_rgb`.

The old behavior can still be obtained by requesting a single variant at a time (no fallback): `mi.set_variant("cuda_ad_rgb")`

## Checklist

### DIFF
--- a/src/python/alias.cpp
+++ b/src/python/alias.cpp
@@ -262,20 +262,22 @@ NB_MODULE(mitsuba_alias, m) {
     mi_dict = m.attr("__dict__").ptr();
     nb::object mi_ext = import_with_deepbind_if_necessary("mitsuba.mitsuba_ext");
     nb::dict mitsuba_ext_dict = mi_ext.attr("__dict__");
-    for (const auto &k : mitsuba_ext_dict.keys())
+    for (const auto &k : mitsuba_ext_dict.keys()) {
         if (!nb::bool_(k.attr("startswith")("__")) &&
             !nb::bool_(k.attr("endswith")("__"))) {
             Safe_PyDict_SetItem(mi_dict, k.ptr(), mitsuba_ext_dict[k].ptr());
         }
+    }
 
     // Import contents of `mitsuba.python` into top-level `mitsuba` module
     nb::object mi_python = nb::module_::import_("mitsuba.python");
     nb::dict mitsuba_python_dict = mi_python.attr("__dict__");
-    for (const auto &k : mitsuba_python_dict.keys())
+    for (const auto &k : mitsuba_python_dict.keys()) {
         if (!nb::bool_(k.attr("startswith")("__")) &&
             !nb::bool_(k.attr("endswith")("__"))) {
             Safe_PyDict_SetItem(mi_dict, k.ptr(), mitsuba_python_dict[k].ptr());
         }
+    }
 
     /// Cleanup static variables, this is called when the interpreter is exiting
     auto atexit = nb::module_::import_("atexit");
--- a/src/python/alias.cpp
+++ b/src/python/alias.cpp
@@ -99,7 +99,7 @@ static void set_variant(nb::args args) {
     for (const auto &arg : args) {
         if (PyDict_Contains(variant_modules, arg.ptr()) == 1) {
             // Variant is at least compiled, we can attempt to use it.
-            valid_variants.append(nb::borrow(arg));
+            valid_variants.append(arg);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/python/alias.cpp
```diff
@@ -116,8 +116,35 @@ static void set_variant(nb::args args) {
...
+        nb::object new_variant_module;
+        try {
+            new_variant_module = variant_module(requested_variant);
+        } catch (const nb::python_error &e) {
+            // The variant failed to import, this could happen e.g. if the
+            // CUDA driver is installed, but there is no GPU available.
+            // We only allow such failures as long as we have more variants to try.
+            if (!is_last && e.matches(PyExc_ImportError)) {
+                // nb::str var_str = requested_variant;
+                // jit_log(LogLevel::Debug,
```
merlinND, line 139: I would have liked to make a debug log here, but `jit_log` is not linked against this module. Is there an easy way to log here?

src/python/alias.cpp
```diff
@@ -95,16 +95,16 @@ static nb::object variant_module(nb::handle variant) {
...
-    nb::object new_variant{};
-    for (auto arg : args) {
-        // Find the first valid & compiled variant in the arguments
+    nb::list valid_variants{};
+    for (const auto &arg : args) {
         if (PyDict_Contains(variant_modules, arg.ptr()) == 1) {
-            new_variant = nb::borrow(arg);
-            break;
+            // Variant is at least compiled, we can attempt to use it.
+            valid_variants.append(nb::borrow(arg));
```
merlinND, line 102: I'm not certain about the ref counting here.

src/python/alias.cpp
njroussel: I see two options:
* We link `libmitsuba.so` to `mitsuba_alias.so` so that we can call `mitsuba::Thread::thread()->logger()->log(level, nullptr, __FILE__, __LINE__, ...)`
* We're guaranteed to have loaded `mitsuba_ext.so`, so we theoretically do a Python call to `mitsuba.Log(Level, message)` through Python.  

I'd personally prefer the second one.

src/python/alias.cpp
njroussel: Fairly certain that the `borrow` is ok. Are there currently any leaks?
I can imagine that a failed `set_variant` itself might leak some stuff too actually.

njroussel (commented):
I like this! I've had it on my backlog to have `mi.variants()` only report valid variants, but there's one nasty problem: if that logic is ever wrong/broken, we'll be skipping hundreds of tests on the CI. Keeping this fallback logic in `set_variant()` shouldn't ever mess with how we've setup the test suite :+1:

src/python/alias.cpp
wjakob: You should not need the borrow at all IMO.

src/python/alias.cpp
merlinND: Thanks, doing a Log call through Python works :+1:

merlinND:
Thanks for the review @njroussel and @wjakob!
I've rebased and incorporated the feedback.

Currently, both on `master` and with this PR, failing to set a variant results in a lot of leaks when exiting, even if the exception is caught and we then successfully set another variant:

```py
# CUDA_VISIBLE_DEVICES= python ./test_failed_variant.py

import mitsuba as mi

if __name__ == "__main__":
    mi.set_log_level(mi.LogLevel.Debug)
    try:
        mi.set_variant("cuda_ad_rgb")
    except ImportError as e:
        print("Failed to set variant")

    print("Current variant:", mi.variant())
    # Still leaks if we then set a valid variant
    mi.set_variant("llvm_ad_rgb")
    print("Current variant:", mi.variant())
```

```
nanobind: leaked 23 instances!
 - leaked instance 0x769448183008 of type "mitsuba.Color0f"
 - leaked instance 0x769443f3bac8 of type "drjit.cuda.ad.TensorXf"
 - leaked instance 0x769443f1ebd0 of type "drjit.scalar.Array3f"
 - ... skipped remainder
nanobind: leaked 273 types!
 - leaked type "drjit.cuda.ad.Int"
 - leaked type "mitsuba.Ray3d"
 - ... skipped remainder
nanobind: leaked 1486 functions!
- ...
```

it's unrelated to this PR, though.

merlinND:
Pinging this PR, would it be okay to merge?

The latest commit ([4306f55](https://github.com/mitsuba-renderer/mitsuba3/pull/1522/commits/4306f55688b1f0deef8828eb8dc15e6600ba1468)) avoids partial initialization of variants when a backend turns out to be unavailable, which used to result in leaks (even before this PR).

njroussel:
LGTM!

Thanks for looking at the leaks too :)

PR #1528, opened 2025-03-20
Improve `Bitmap` support in Jupyter notebooks

wjakob:
The `Bitmap._repr_html()` enables viewing Mitsuba `Bitmap` objects in Jupyter notebooks. This commit makes the following changes:

- Use JPG (compression level 99%) instead of PNG images, which is faster to de/encode.

- Base-64 encode locally instead of copying back and forth several times.

- Don't force the image size to 300 px, just give the image and let the Jupyter notebook figure out how to display it.

### DIFF
--- a/include/mitsuba/core/platform.h
+++ b/include/mitsuba/core/platform.h
@@ -20,11 +20,13 @@
 #  define MI_IMPORT   __declspec(dllimport)
 #  define MI_NOINLINE __declspec(noinline)
 #  define MI_INLINE   __forceinline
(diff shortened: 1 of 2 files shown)


njroussel (commented):
LGTM

My only concern was how this would affect the current tutorials. Of course, they won't change on [mitsuba.readthedocs.io](mitsuba.readthedocs.io) unless we re-run them. I tried a few of them and the layout didn't change too much :+1: .

njroussel (approved):

PR #1563, opened 2025-04-11
Simplified OptiX intersection programs

wjakob:
The OptiX intersection programs for various shapes contained an
unnecessary world <-> object space transformation. This step was not
only unnecessary, but it generates a shockingly large amount of code
related to OptiX motion transforms and all sorts of features that we are
to using at all. All of this logic consumes registers, which has a
negative impact on occupancy of generated kernels.

This commit incorporates a small change to the `get_ray()` function and
adapts the other intersection programs, which brings `optix_rt.ptx` to
12% of its current size.


wjakob:
Hooray, it passes the test suite 🎉

PR #1565, opened 2025-04-11
Fix zeroing of OptiX structs

dvicini:
Similar to https://github.com/mitsuba-renderer/drjit-core/pull/138, this fixes the zero initialization of structs with union members. In particular, it prevents segfaults due to incomplete initialization of the OptixProgramGroupDesc structs.

The issue was uncovered by running with a sanitizer that explicitly does not zero initialize memory

### DIFF
--- a/src/render/scene_optix.inl
+++ b/src/render/scene_optix.inl
@@ -236,7 +236,8 @@ size_t init_optix_config(uint32_t optix_config_shapes) {
     // program groups will have a closest hit program.
 
     OptixProgramGroupOptions program_group_options = {};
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @dvicini ,

Nicolas just redesigned the OptiX interface rather significantly. Is this commit still applicable? If so, could you rebase it to the current version?

dvicini:
Thanks for letting me know. I rebased and it seems that the commit is still applicable. By zeroing the structs explicitly, some of the code can be slightly simplified. Moreover, the current, updated master branch incompletely zeros out the pointers in the struct, making the bug even more subtle than it used to be.

wjakob:
Thanks!

PR #1570, opened 2025-04-16
Re-enable Python bindings for Properties::Type (fixes the Blender plugin)

dvicini:
This PR re-enables the Python bindings for Properties::Type. The motivation for this is that the Blender importer queries the types of properties (e.g., https://github.com/mitsuba-renderer/mitsuba-blender/blob/5a12adda2d0da3ccf3c125a54e36f7cf57988957/mitsuba-blender/io/importer/emitters.py#L52), which currently is broken.

One minor complication is that the enum is nested in Properties, which itself is variant dependent. However, the enum isn't. I therefore added some logic to detect if the type was already bound, similar to the `MI_PY_CHECK_ALIAS` macro. I am not sure if this is the best option, but it seems to work.

I also modified the unit tests to exercise the functionality

### DIFF
--- a/src/core/python/properties_v.cpp
+++ b/src/core/python/properties_v.cpp
@@ -153,23 +153,24 @@ MI_PY_EXPORT(Properties) {
             .def(nb::self != nb::self, D(Properties, operator_ne))
             .def_repr(PropertiesV);
 
-        // FIXME: Binding this enumeration leaks. Defining an internal enum to
-        // an arbitrary class is fine, this seems to be specifically an issue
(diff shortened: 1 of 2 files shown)


wjakob:
I suspect the leaks are a remnant from an older version of nanobind, where enumerations were done with `nb::class_<>`, which often caused leaks due to circular references. The change looks good, I will merge it. Unrelated: there is some refactoring to `Properties` planned (`refactor` branch) that will likely need some careful attention to the Blender plugin to avoid breaking things.

PR #1581, opened 2025-04-23
Reduce synchronizations when setting up scene in OptiX 

njroussel:
## Description

This PR removes some unnecessary synchronizations that were being explicitly triggered during the OptiX scene setup.
In addition, it also removes some superfluous calls to `dr::eval()`.

Notes:
Most shapes explicitly synchronize on a call to `parameters_changed()`. This is theoretically only necessary on LLVM backends, as the scalar values are used in the ray intersection routines (by synchronizing, we wait for any running kernels to terminate before updating the scalar parameters). However, in practice, the shape parameters are stores as `field` objects and an update of the JIT value will trigger a synchronization to keep the scalar value up-to-date too.  In another PR, I will try to alleviate this problem.

## Testing

No new tests.

### DIFF
--- a/include/mitsuba/render/optix/shapes.h
+++ b/include/mitsuba/render/optix/shapes.h
@@ -133,13 +133,16 @@ void build_gas(const OptixDeviceContext &context,
     // Build a GAS given a subset of shape pointers
     auto build_single_gas = [&context](const std::vector<ref<Shape>> &shape_subset,
                                        MiOptixAccelData::HandleData &handle) {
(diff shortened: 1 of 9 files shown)


wjakob:
Looks good in general, though the failing testcase seems concerning.

njroussel:
The failing test is unrelated to these changes. It's been happening since the `rectangle` PR, I imagine that the silhouette sampling isn't as smooth anymore, the tolerances of the test must be updated. I'll have a look at that in a different commit/PR.

PR #1619, opened 2025-06-06
(prb)volpath: fix hero-wavelength calculation in volumetric integrators

lnuic:
TODO


njroussel:
Thanks, replaced by #1628

PR #1622, opened 2025-06-10
Remove `Thread`, `ThreadEnvironment`, `ScopedSetThreadEnvironment` classes

wjakob:
Everything related to multithreading in Mitsuba is nowadays done by the nanothread thread pool. This commit removes the ``Thread``, ``ThreadEnvironment``, ``ScopedSetThreadEnvironment`` classes that were remainders of the Mitsuba 1/2 architecture.

Only a few `Thread` members remain for reasons of compatibility (but with deprecation warnings):

- ``Thread.file_resolver()``, ``Thread.logger()``, ``Thread.thread()``, ``Thread.wait_for_tasks()``.

The class will be entirely removed in a future release.

This commit also removes thread name printing from logger class.

### DIFF
--- a/src/core/progress.cpp
+++ b/src/core/progress.cpp
@@ -52,7 +52,7 @@ void ProgressReporter::update(float progress) {
         memcpy((char *) m_line.data() + eta_pos, eta.data(), eta.length());
     }
 
-    mitsuba::logger()->log_progress(progress, m_label, m_line, eta, m_payload);
+    logger()->log_progress(progress, m_label, m_line, eta, m_payload);
     m_last_update = elapsed;
 }
 
--- a/src/core/xml.cpp
+++ b/src/core/xml.cpp
@@ -602,7 +602,7 @@ static std::pair<std::string, std::string> parse_xml(XMLSource &src, XMLParseCon
                     check_attributes(src, node, { "value" });
                     if (depth != 1)
                         src.throw_error(node, "<path>: path can only be child of root");
-                    ref<FileResolver> fs = mitsuba::file_resolver();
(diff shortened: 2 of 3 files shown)


wjakob:
cc @dvicini @merlinND @njroussel

wjakob:
(Claude code did a lot of the grunt work on this one. Pretty cool!)

dvicini:
This is great, I like that this removes the issue with the shutdown of static thread local objects

src/core/python/logger.cpp
```diff
@@ -60,4 +58,6 @@ MI_PY_EXPORT(Logger) {
         .def_method(Logger, read_log);
 
     m.def("Log", &PyLog, "level"_a, "msg"_a);
+    m.def("set_logger", &mitsuba::set_logger, "logger"_a);
```
merlinND, line 61: If I remember correctly, `.none()` needs to be added to the argument to make it explicitly allowed?
From the PR it looks like have no logger is allowed, so we could pass None here to do that.

src/core/logger.cpp
```diff
@@ -65,7 +65,7 @@ void Logger::log(LogLevel level, const Class *class_, const char *file,
     }
 
     std::string text = d->formatter->format(level, class_,
-        Thread::thread(), file, line, msg);
+        file, line, msg);
 
     std::lock_guard<std::mutex> guard(d->mutex);
```
merlinND, line 70: I didn't realize this was already guarded by a mutex, so actually there's nothing new to do now that the logger is process-wide by default.

merlinND (commented):
Nice simplification! Always good to get rid of unnecessary code.

PR #1630, opened 2025-06-18
Cleanup pass over the codebase to remove technical debt

wjakob:
This large commit modernizes Mitsuba 3's core infrastructure. It removes legacy code dating back to Mitsuba 0.6.x and streamlines the rest of the codbase to adapt to these changes.

It includes the following notable changes:

- **Run-time type information (RTTI)**: Mitsuba previously shipped a home-grown RTTI solution implemented by ``mitsuba::Class``. This PR replaces it with standard C++ features such as ``dynamic_cast`` and ``typeid()``. Object subclasses implement ``Object::type()``, which performs a coarse-grained classification into major plugin categories. The use of the ``MI_IMPLEMENT_CLASS()`` macro was removed following this change, as it is no longer needed.

  The `Object::class_()` member no longer exists. To get the class name of an instance, call `.class_name()`, the variant name is available via `.variant_name()`.

- **Properties**: this commit adds an all-new implementation of the ``Properties`` class that is used to pass parameter values to plugins. The previous version built on top of a custom variant data type dating back to C++11, where `std::variant<..>` did not yet exist, and it used a complex scheme to share property storage between generic and variant-specific Properties objects in Python.

  The new version avoids this complexity and delegates most work to ``std::variant<T>``. It adds a type-safe way of passing arbitrary objects between plugins (e.g., tensors) that are encoded in a ``std::any`` like class named ``mitsuba::Any``, and which enables seamless use in Python bindings.

  Previously, there were many type-specific ``Props::set_X`` functions that are now all subsumed by a generic ``Props::set<X>``, similar to how this has worked already for getters.

  The Python bindings of the ``Properties`` class provide a better approximation of a Python dictionary.

   - `props.has_property(key)` -> `key in props`
   - `props.remove_property(key)` -> `del props[key]`
   - `props.property_names()` -> `property.keys()`

  There is also `property.items()`. The three methods above still exist
  but trigger a deprecation warning.

- **Plugin loading**: the ``PluginManager`` class was rewritten and simplified. The new versions adds cleanup hooks to properly support garbage collection of Python-based extensions. Calls to ``dr::eval()`` (important for kernel caching in parallel usage) were moved into the plugin manager.

- **String calling conventions**: Mitsuba interfaces that take non-owning references to strings as an input previously did so inconsistently either via ``const std::string &`` or ``const char *``. I changed many of those to take ``std::string_view`` instead, which is compatible with both kinds of inputs.

- **AD integrator initialization**: Mitsuba explicitly reload the AD integrators when changing variants, but this was happening even during the initial import. The commit changes this to only do a single import.

- **Variant objects**: Scene objects now inherit from ``JitObject<Derived>``, which is responsible for identifying the variant, and potentially registering the instance with the Dr.Jit instance registry. The macro ``MI_REGISTRY_PUT()`` was removed globally. The variant object also provides a ``Variant`` member containing the variant name, which is determined using a partial template overload (rather than a constexpr function).

  Another consequence of the variant-related changes is that all ``MI_CALL_*`` macros could be replaced with the more generic Dr.Jit ``DRJIT_CALL_*`` macros.

- Scene traversal: Mitsuba plugin classes implement a `.traverse()` method that enables recursive traversal of members. This involved many lines of the form `callback->put_parameter(...)` or `callback->put_object()`. This is now shortened to `cb->put()`, subsuming both cases. The distinction between sub-objects and array parameters is detected using the type system. One can now directly add `ref<T>` members (no `.get()` needed) and the device portion part of `field<T>` members, which unifies the syntax. It is no longer necessary to convert ParamFlags into integers using syntax like `+ParamFlags::Differentiable`.

  The Python bindings of the scene traversal and SceneParameters object previously used an implementation that tried to brute force-convert the type to a set of supported types one by one until success. This was replaced by a more efficient direct implementation using the nanobind low level interface.

- **Minor**: Removed the ``AtomicFloat`` class.

User C++ Mitsuba extensions will require touch-ups to adapt to the macro and ``Properties``-related changes.


wjakob:
CC @dvicini @merlinND @Speierers

PR #1635, opened 2025-06-21
Clean up test_freeze's file paths

dvicini:
Minor cleanup to make `test_freeze` use files from the `resources/tests` and `resources/common` folders, instead of the tutorials and documentation.


wjakob:
Thanks!

PR #1641, opened 2025-06-24
Bugfix Sunsky emitter updates

matttsss:
## Description

This PR fixes an issue where the Sunsky plugin's datasets were re-computed even though the updated parameter did not influence it. 
More precisely:
* The datasets were re-computed when the changed keys were empty.
* If the above did not occur, the macro "function" `CHANGED` evaluated to `true` only if `albedo` was changed.
* The TGMM data would not update if the sun's direction was changed via the `sun_direction` parameter.

## Testing

This PR also adds a test in `test_sunsky.py` that updates the plugin's parameters to make sure the changes are correctly registered.

## Checklist

### DIFF
--- a/src/emitters/sunsky.cpp
+++ b/src/emitters/sunsky.cpp
@@ -264,13 +264,13 @@ public:
 
         #define CHANGED(word) string::contains(keys, word)
 
-        bool changed_atmosphere = CHANGED("albedo") || CHANGED("turbidity");
-        bool changed_time_record = !keys.empty() && m_active_record && (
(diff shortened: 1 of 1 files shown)


njroussel (changes requested):
The new tests look good to me. 

However, it's by design that plugins should entirely update themselves if the set of keys is empty (see [docs](https://github.com/mitsuba-renderer/mitsuba3/blob/master/include/mitsuba/core/object.h#L135-L136))

matttsss:
Right, sorry for not spotting that, it should be good now @njroussel

PR #1643, opened 2025-06-25
Fix use of DRJIT_CALL_END macro in mesh.h

dvicini:
Minor: fix use of call end macro in mesh.h to use the correct "inherited" version


dvicini:
nevermind - I saw there are some DrJit changes on a new branch that fix this

PR #1655, opened 2025-07-04
Ray loader to support mini-batch training

ziyi-zhang:
This PR introduces a ray loader that wraps a list of sensors to enable mini-batch optimization.
The sampling strategy aims to uniformly cover all pixels before reshuffling (i.e., permutation-based sampling), but this is not strictly guaranteed when a tile size is specified.

## TODO

- [x] Tiling for GPU coherence.
- [x] FP16 support (FP16 will not be supported as mi.render by default forces float32)
- [x] Add tests
- [x] Support ptracer & projective: don't sample pixels, sample a batched K source sensors, where K is the smallest number to saturate `pixels_per_batch` (Not supported for now).


merlinND:
Hi @ziyi-zhang,

I haven't ready the implementation so maybe this is already supported, but one feature that could be quite helpful: ensure that rays can be sampled as e.g. 2x2, 4x4, etc tiles from the same sensor. The user could select the tile size to get more or less coherence.

ziyi-zhang:
Hi @merlinND,

Good point.

What do you think is the best solution if the film size is not divisible by the patch size, e.g., 4x4 patch for a 12x10 film?
A related question is the division of `pixels_per_batch` vs. `total_pixels`. I currently pad total reference pixels with random pixels (so some pixels may appear twice) to make sure it's a multiple. But it may be better to just pad with dummy pixels?

merlinND:
Hi @ziyi-zhang,

I haven't checked the implementation, but it sounds like you are trying to exhaustively enumerate all batches (and maybe now all NxN ray tiles) out of the total pixels, without overlap.
I believe it would be fine to instead sample batches / tiles _with overlap allowed_, which means the tiles / batches just need to fit in the ref image dimensions, rather than being divisible.

ziyi-zhang:
Set tile size to 4 by default.

- Very large tile sizes are bad.
- Tile size not a multiple of 2 is bad.

dual_axis_comparison

src/python/python/ad/loaders/ray_loader.py
```diff
@@ -0,0 +1,469 @@
+from __future__ import annotations as __annotations__ # Delayed parsing of type annotations
+
+from typing import Tuple
+
+import mitsuba as mi
+import drjit as dr
+
+
+class flat_sensor(mi.Sensor):
```
wjakob: CamelCase for classes

src/python/python/ad/loaders/ray_loader.py
```diff
@@ -0,0 +1,469 @@
...
+                '    height=%d,\n'
+                ']' % (self.num_sensors,
+                       self.source_film_width,
+                       self.source_film_height))
+
+mi.register_sensor("flat_sensor",
+                   lambda props: flat_sensor(props))
+
+
+class Rayloader():
```
wjakob: RayLoader?

src/python/python/ad/loaders/ray_loader.py
```diff
@@ -0,0 +1,469 @@
...
+    def __init__(
+        self,
+        sensors: list[mi.Sensor] | mi.Sensor,
+        target_images: list[mi.TensorXf] | mi.TensorXf,
```
wjakob, line 155: Needs docstrings for all arguments., more detailed documentation.

src/python/python/ad/loaders/ray_loader.py
wjakob: Class docsring. What does this do?

src/python/python/ad/loaders/ray_loader.py
```diff
@@ -0,0 +1,469 @@
...
+            raise ValueError(
+                "Target images must have shape (height, width, channels)"
+            )
+
+        reference_film = sensors[0].film()
+        self.num_sensors = len(sensors)
+        self.width = reference_film.size()[0]
+        self.height = reference_film.size()[1]
+        self.channel_size = reference_shape[2]
+        self.ttl_pixels = self.num_sensors * self.width * self.height
```
wjakob: Describe members. ttl_pixels? unclear.

src/python/python/ad/loaders/ray_loader.py
wjakob: rename RayDataLoader (also elsewhere in docs)

src/python_tests/test_ray_loader.py
```diff
@@ -0,0 +1,639 @@
...
+    opt[key] = params[key]
+    params.update(opt)
+
+    # Render reference images with target reflectance (0.8)
+    scene_dict['shape']['bsdf']['reflectance'] = 0.8
+    target_scene = mi.load_dict(scene_dict)
+    image_ref = []
+    for sensor in sensors:
+        image_ref.append(mi.render(target_scene, sensor=sensor, spp=4))
+
```
wjakob, line 591: Will need a small tutorial on how to use.

wjakob (commented):
Some feedback

PR #1656, opened 2025-07-05
Attempt removing GC calls from the test suite

wjakob:
I've noticed when running small self-contained tests that a significant portion of the runtime is spent in the cleanup callback after each executed test run. (A set of 20 tests went from ~6s to .04s after removing the GC calls). Let's see what happens at a larger scale when removing this..


wjakob:
Looks like we don't need them :)

PR #1664, opened 2025-07-08
Add back set_id method and test its use in load_dict

dvicini:
Adds back the `set_id` method and calls it during `load_dict`. This makes it such that plugins that already were instantiated and then used in a dict get the ID of the dictionary key (reproducing the previous behavior)


wjakob:
LGTM, thanks!

Issue #1649, opened 2025-07-01
Deadlock in load_dict, with loglevel >= Debug


DoeringChristian:
Hi,

# Summary

When loading a simple scene such as the cornell box Mitsuba3 freezes (deadlocks) on the current master branch, when `loglevel >= Debug.

# Details

Loading a scene in parallel i.e. with `dr.load_dict(dict, parallel=True)`, when the log level is set to anything more detailed than `Info`, can cause a deadlock in `jit_new_scope`, called from `instantiate_node`. Locking the mutex in this function seems to conflict with another thread, locking the state mutex for logging. The other lock seems to be taken inside the `log_callback`, called from `jitc_log`, as commenting out this call resolves the deadlock.

The commit I tested on is `60cd59b8`, which already includes #1622.
I am running Ubuntu, and compiled Mitsuba3 with gcc13.2.0 (conda-forge).

This is a simple reproducer:
```python
import mitsuba as mi
import drjit as dr

mi.set_variant("cuda_ad_rgb")

dr.set_log_level(dr.LogLevel.Debug)

scene = mi.load_dict(mi.cornell_box(), parallel=True)
```

What I have already tried:
- Reverting the spinlock logic in `drjit-core:src/lock.h` to the pre-freezing feature state did not resolve the deadlock
- Commenting out line 55 in `drjit-core:src/log.cpp` resolved the deadlock
- Setting `parallel=False` resolved the deadlock
- The branch already includes #1622 

Possibly related to #1572.

njroussel:
Addressed in #1671

PR #1671, opened 2025-07-11
deadlock bugfix for old parser

wjakob:
Attempted fix for a deadlock issue that specifically affects the old parser. I am not sure why this came up only now.

### DIFF
--- a/src/core/python/xml_v.cpp
+++ b/src/core/python/xml_v.cpp
@@ -125,7 +125,10 @@ MI_PY_EXPORT(xml) {
             try {
                 parse_dictionary<Float, Spectrum>(ctx, "__root__", dict);
                 std::unordered_map<std::string, Task*> task_map;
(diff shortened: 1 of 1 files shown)


merlinND:
Attempts to fix https://github.com/mitsuba-renderer/mitsuba3/issues/1649.

merlinND:
Would it be worth adding the reproducer from @DoeringChristian as a unit test?
Of course deadlocks in the CI are not ideal, but at least the problem would be highly visible if it shows up again.

Something like:
```py
def test_logger_deadlock():
    log_level = dr.log_level()
    dr.set_log_level(dr.LogLevel.Debug)
    scene = mi.load_dict(mi.cornell_box(), parallel=True)
    dr.set_log_level(log_level)
```

wjakob:
Ok, sure -- I will commit it separately.

PR #1676, opened 2025-07-14
Remove the old XML parser/writer

wjakob:
Following PR #1669, this PR deletes the old XML parser and testcases.


wjakob:
I had forgotten `xml_v.cpp` in this PR. The full stats are:
```17 files changed, 145 insertions(+), 4585 deletions(-)```

PR #1679, opened 2025-07-15
Refactoring pass over shape and sensor transformations

wjakob:
This PR bundles three related changes:

**Refactored homogeneous coordinate transformations** (``Transform3f``, ``Transform4f``, etc.)

Mitsuba uses a generic 4x4 matrix class to represent various types of transformations, such as rotations, translations, scales, and potentially perspective effects. The class also internally stores a precomputed inverse transpose, which is necessary to efficiently transform surface normals.

However, this representation ends up being rather heavy. A 4x4 branchless matrix inverse (needed for the inverse transpose) computed via Cramer's rule expands into a *huge* computation graph in the JIT/AD backend. In practice, perspective transformations are just used in two places: `perspective.cpp` and `thinlens.cpp`, and a simpler affine representation can be used everywhere else.

This PR therefore splits ``Transform`` into two variants (selected by a template parameter): ``AffineTransform`` and ``ProjectiveTransform``. The former uses several optimizations : cheaper point transformations, cheaper composition of transforms, and cheaper precomputation of the inverse transpose. There are no storage savings (a full 4x4 matrix is still used). However, the JIT will optimize away the unused parts in practice, and this is expected to work more reliably following this PR.

C++ code will need to be updated to reference either ``AffineTransform`` or ``ProjectiveTransform``. The ``Properties`` class only stores projective transformations. Python has bindings for both flavors, with the `mi.Transform*` providing aliases to `mi.AffineTransform*`.

**Removed redundant transformations**

All shapes stored two redundant transformations: ``m_to_object``, and ``m_to_world``. Several sensors furthermore stored redundant transformations: ``m_camera_to_sample`` and ``m_sample_to_camera``. They are inverses of the other and trivial to obtain, since the class already stores the inverse transpose. The PR removes these and uses `.inverse()` to obtain the inverse form as needed.

**Simpler update pattern**

The ``parameters_changed()`` method of a number of plugins contained the following pattern to update & differentiably recompute the `inverse_transpose` member of ``m_to_world``:

```cpp
if constexpr (dr::is_diff_v<Float>) {
    AffineTransform4f to_world = m_to_world.value();
    // Re-attach inverse_transpose to original matrix
    if (dr::grad_enabled(to_world.matrix)) {
        Matrix4f invt_diff = dr::inverse_transpose(to_world.matrix);
        to_world.inverse_transpose =
            dr::replace_grad(to_world.inverse_transpose, invt_diff);
    }
    m_to_world = to_world;
} else {
    // Update the scalar value of the matrix
    m_to_world = m_to_world.value();
}
```

This is now replaced by:
```cpp
m_to_world = m_to_world.value().update();
```

### DIFF
--- a/src/integrators/tests/test_ad_integrators.py
+++ b/src/integrators/tests/test_ad_integrators.py
@@ -561,7 +561,7 @@ class TranslateOccluderAreaLightConfig(TranslateShapeConfigBase):
         }
         self.ref_fd_epsilon = 1e-3
         self.error_mean_threshold = 0.03
-        self.error_max_threshold = 1.65
+        self.error_max_threshold = 1.75
         self.error_mean_threshold_bwd = 0.25
         self.integrator_dict = {
             'max_depth': 2,


wjakob:
1. Computation graph when initializing a `Transform4f` from an opaque matrix (old implementation):
4x4 inverse

2. Computation graph when initializing an `AffineTransform4f` from an opaque matrix (new implementation):  affine

include/mitsuba/core/properties.h
```diff
@@ -17,7 +17,7 @@ template <typename T, typename = void> struct prop_map { using type = void; };
 template <typename T> using prop_map_t = typename prop_map<std::decay_t<T>>::type;
 template <typename T> struct base_type;
 template <typename T> struct is_transform_3: std::false_type { };
-template <typename T> struct is_transform_3<Transform<Point<T, 3>>> : std::true_type { };
+template <typename T> struct is_transform_3<AffineTransform<Point<T, 3>>> : std::true_type { };
```
merlinND, line 20: According to the PR description, shouldn't this be `ProjectiveTransform`?
Properties should use the most general / powerful transform type I think?

include/mitsuba/core/transform.h
```diff
@@ -52,45 +60,66 @@ template <typename Point_> struct Transform {
...
-    MI_INLINE Transform operator*(const Transform &other) const {
-        return Transform(matrix * other.matrix,
-                         inverse_transpose * other.inverse_transpose);
-    }
-
-    /// Compute the inverse of this transformation (involves just shuffles, no arithmetic)
     MI_INLINE Transform inverse() const {
-        return Transform(dr::transpose(inverse_transpose), dr::transpose(matrix));
+        return Transform(dr::transpose(inverse_transpose),
+                         dr::transpose(matrix));
```
merlinND, line 84: In all classes where we had stored both directions (e.g. `m_camera_to_sample`, `m_sample_to_camera`), are these additional 2 `dr::transpose()` calls every time the field is used worth it?

include/mitsuba/core/transform.h
```diff
@@ -104,90 +133,6 @@ template <typename Point_> struct Transform {
         return dr::all_nested(matrix != t.matrix);
     }
 
-    /**
-     * \brief Transform a 3D vector/point/normal/ray by a transformation that
-     * is known to be an affine 3D transformation (i.e. no perspective)
-     */
-    template <typename T>
-    MI_INLINE auto transform_affine(const T &input) const {
```
merlinND, line 112: Should we maybe keep this with a deprecation warning? (like in Python)

include/mitsuba/core/transform.h
```diff
@@ -269,9 +177,9 @@ template <typename Point_> struct Transform {
      * \param up     Up vector
      */
     template <size_t N = Size, dr::enable_if_t<N == 4> = 0>
-    static Transform look_at(const Point<Float, 3> &origin,
-                             const Point<Float, 3> &target,
-                             const Vector<Float, 3> &up) {
+    static Transform look_at(const mitsuba::Point<Float, 3> &origin,
+                            const mitsuba::Point<Float, 3> &target,
+                            const Vector<Float, 3> &up) {
```
merlinND, line 182:
```suggestion
    static Transform look_at(const mitsuba::Point<Float, 3> &origin,
                             const mitsuba::Point<Float, 3> &target,
                             const Vector<Float, 3> &up) {
```

include/mitsuba/core/properties.h
wjakob: We never pass projective transformations to plugins

include/mitsuba/core/transform.h
wjakob: For the JIT, those don't matter. For scalar mode, there might be a cost (but I would expect this to be rather negligible).

include/mitsuba/core/transform.h
wjakob: Fair enough.

src/core/tests/test_transform.py
```diff
@@ -93,99 +120,137 @@ def check_inverse(tr, expected):
...
     assert T.look_at([0, 0, 0], [0, 1, 0], [1, 0, 0]).scale(4.0) == T.look_at([0, 0, 0], [0, 1, 0], [1, 0, 0]) @ T.scale(4.0)
     assert T.from_frame(mi.Frame3f([1, 0, 0])).scale(4.0) == T.from_frame(mi.Frame3f([1, 0, 0])) @ T.scale(4.0)
     assert T.to_frame(mi.Frame3f([1, 0, 0])).scale(4.0) == T.to_frame(mi.Frame3f([1, 0, 0])) @ T.scale(4.0)
 
 
-def test06_dual_callable(variant_scalar_rgb):
-    """Test dual callable functionality - calling Transform methods as class methods"""
-    # Dual callable is automatically enabled when Transform classes are created
+@transform_types
+def test06_dual_callable(variant_scalar_rgb, transform_type):
```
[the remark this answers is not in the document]
wjakob: Will do.

PR #1680, opened 2025-07-17
AD integrators: differentiating geometry

### LINKED ISSUE #1506: PRB gradients biased if scene is not static
Following our private exchange earlier last month, it's probably good to have an official tracker for this issue, also so that users making similar observations in their applications have a reference.

## Summary

There is a theoretical gap in the combination of Radiative Backpropagation/Path Replay Backpropagation and moving geometry because the original papers (implicitly) assumed a static scene. An immediate consequence is that all PRB integrators produce biased gradients when a parameter affects geometry, including the deprecated `prb_reparam` and the current `prb_projective` integrators:

A central point is that when geometry moves, detached sampling of (hemi)sphere directions and detached sampling of surface points need significantly different approaches: the former would now need to track global path information and require an implementation akin to attached sampling; the latter is local but one must ensure to sample surface points (e.g. intersect with FollowShape) and consider the (derivative) of the transformation's Jacobian determinant.

[This repository](https://github.com/mworchel/mitsuba-missing-term) bundles code to reproduce the issue and an implementation of integrators that solve the problem. The theoretical side is covered in a technical report that we will probably make public next month (I will insert a link here).

njroussel:
## Description

This PR updates all AD integrators to correctly handle missing _continuous_ derivative terms arising from moving geometry. 

Mitsuba's integrators follow the methods described in [RB](https://rgl.epfl.ch/publications/NimierDavid2020Radiative) and [PRB](https://rgl.epfl.ch/publications/Vicini2021PathReplay), which focus on differentiating scenes with _static_ geometry. 
This assumption is to exclude [boundary derivatives](https://rgl.epfl.ch/publications/Zhang2023Projective), which are a separate concern and handled independently.
However, moving geometry introduces non-boundary derivative terms that were previously unaccounted for in Mitsuba 3's AD integrators.
This PR re-attaches these missing terms.

See #1506 and [this paper](https://diglib.eg.org/handle/10.2312/sr20251198) for more context. We thank the authors for spotting and investigating the issue.

Although `prb` and `prb_basic` were updated, they still do not compute the boundary integral and are therefore still typically ill-suited for geometry optimization tasks. 

## Performance

**When the scene's geometry is static, the updated integrators bare no additional runtime cost.** Dr.Jit can prune the new terms as they will not be attached (AD) for static geometry.

The new integrators required some extra changes to fully benefit from SER. Unexpectedly, the **new integrators reduce the average runtime by ~30% when compared to the older ones** (primal evaluation). Both versions shuffle threads based on the shape ID of the current surface interaction. However, the old version also used some encoding of the XYZ coordinates of the hitpoint. I believe that this extra information effectively scheduled all warps to be shuffled in the hope of improving data access coherence. On the other hand, by exclusively using the shape ID, the new integrators might have warps that will skip the SER process entirely if the warp is coherent. (This is just a hypothesis.)

Here's a summary of the speedups I've benchmarked with `prb` on some standard scenes:
| scene name     | speedup   |
|:---------------|:----------|
| staircase      | 1.42x     |
| living-room    | 1.25x     |
| bathroom | 1.31x     |
| spaceship      | 1.23x     |

We should maybe apply this change to the non-AD integrators of Mitsuba too.

Fixes #1506

## Testing

Five new configurations were added to `test_ad_integrators.py`, all of which almost exclusively produce gradients from the missing terms. I've also spent some time to tighten the thresholds on all configurations. Some configurations are particularly expensive, I might update their resolution/spp if I see a noticeable impact on the CI.

:warning: The `resources/data` submodule changed should be merged to `mitsuba-data:master` before this PR is merged.

### DIFF
--- a/src/emitters/projector.cpp
+++ b/src/emitters/projector.cpp
@@ -223,7 +223,7 @@ public:
         ds.n       = m_to_world.value() * ScalarVector3f(0, 0, 1);
         ds.uv      = uv;
         ds.time    = it.time;
-        ds.pdf     = 1.f;
+        ds.pdf     = dr::select(active, 1.f, 0.f);
         ds.delta   = true;
         ds.emitter = this;
         ds.d       = ds.p - it.p;
@@ -279,6 +279,13 @@ public:
         SurfaceInteraction3f it_query = dr::zeros<SurfaceInteraction3f>();
         it_query.wavelengths = it.wavelengths;
         it_query.uv = ds.uv;
+        if constexpr (dr::is_diff_v<Float>) {
+            // Re-attach UV
+            if (dr::grad_enabled(it_local) && !dr::grad_enabled(ds.uv)) {
+                Point2f uv_diff = dr::head<2>(m_sample_to_camera.inverse() * it_local);
+                it_query.uv = dr::replace_grad(it_query.uv, uv_diff);
+            }
+        }
 
         UnpolarizedSpectrum spec = m_irradiance->eval(it_query, active);
         spec *= dr::Pi<Float> * m_intensity_scale /
--- a/src/emitters/projector.cpp
+++ b/src/emitters/projector.cpp
@@ -280,7 +280,8 @@ public:
         it_query.wavelengths = it.wavelengths;
         it_query.uv = ds.uv;
         if constexpr (dr::is_diff_v<Float>) {
-            // Re-attach UV
+            // Re-compute attached UV when the shading point is moving and the
+            // input ds.uv has not been re-computed
             if (dr::grad_enabled(it_local) && !dr::grad_enabled(ds.uv)) {
                 Point2f uv_diff = dr::head<2>(m_sample_to_camera.inverse() * it_local);
                 it_query.uv = dr::replace_grad(it_query.uv, uv_diff);
--- a/src/integrators/tests/test_ad_integrators.py
+++ b/src/integrators/tests/test_ad_integrators.py
@@ -146,7 +146,7 @@ class ConfigBase:
     requires_discontinuities = False
 
     def __init__(self) -> None:
-        self.spp = 128
+        self.spp = 1024
         self.res = 128
         self.error_mean_threshold = 0.05
         self.error_max_threshold = 0.5
@@ -187,7 +187,7 @@ class ConfigBase:
 
         @fresolver_append_path
         def create_scene():
-            return mi.load_dict(self.scene_dict)
+            return mi.load_dict(self.scene_dict, optimize=False)
         self.scene = create_scene()
         self.params = mi.traverse(self.scene)
 
@@ -210,6 +210,7 @@ class ConfigBase:
                f'  spp = {self.spp},\n' \
                f'  key = {self.key if hasattr(self, "key") else "None"}\n' \
                f']'
+
 
 # BSDF albedo of a directly visible gray plane illuminated by a constant emitter
 class DiffuseAlbedoConfig(ConfigBase):
@@ -238,8 +239,6 @@ class DiffuseAlbedoConfig(ConfigBase):
 class DiffuseAlbedoGIConfig(ConfigBase):
     def __init__(self) -> None:
         super().__init__()
-        self.spp = 128
-        self.error_mean_threshold = 0.06
         self.key = 'green.bsdf.reflectance.value'
         self.scene_dict = {
             'type': 'scene',
@@ -307,7 +306,6 @@ class AreaLightRadianceConfig(ConfigBase):
 class DirectlyVisibleAreaLightRadianceConfig(ConfigBase):
     def __init__(self) -> None:
         super().__init__()
-        self.spp = 128
         self.key = 'light.emitter.radiance.value'
         self.scene_dict = {
             'type': 'scene',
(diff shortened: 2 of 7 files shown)


ziyi-zhang:
## Technical details

The continuous/shading derivatives of a light path arising from one interaction point can be obtained by differentiating a three-point form (in vague terms, also see [Appendix 1.1 here](https://arxiv.org/pdf/2408.16005)):



where RB and PRB mainly differentiates BSDF properties and normals.

When the geometry moves, differentiating the above formulation requires a parameterization---specifically, how the interaction point (e.g., $\mathbf{x}_\mathrm{k}$) depends on the parameter $\pi$. There are multiple valid parameterizations that can be used. While all of them yield unbiased derivative estimates, they differ in terms of variance, implementation complexity, and practical limitations. This is conceptually similar to the distinction between [detached and attached estimators](https://rgl.epfl.ch/publications/Zeltner2021MonteCarlo).

We mainly explored the following two parameterizations: 
(a) Treat  $\mathbf{x}_\mathrm{k}$  as glued on the surface. This corresponds to `mi.RayFlags.FollowShape` in `compute_surface_interaction()`;
(b) Treat  $\mathbf{x}_\mathrm{k}$  as the result of a ray-shape intersection, and other points as static (to avoid recursive dependency). 



We observed the following differences between the two parameterizations:
- Option (a) exhibits slightly lower variance in a fully diffuse Cornell box, though the difference is marginal.
- Option (b) consistently achieves lower variance in the lens test across all roughness levels. This scene consists of three rough dielectric lenses aligned in a row and contains no discontinuities. Notably, as the lens surface becomes shinier (i.e., lower roughness), the variance of option (a) increases significantly. This behavior is expected, as option (b) preserves the BSDF value at $\mathbf{x}_{k-1}$, thereby avoiding additional variance injection from that term.
- Option (a) is incompatible with a non-smooth pixel reconstruction filter (e.g., `box`). Option (b) does not differentiate the camera ray direction and is thus free of this limitation.
- Option (a) differentiates terms that depend on a local UV parameterization, whose quality can directly impact variance.
- Option (a) is approximately 50% slower than option (b), based on measurements that exclude discontinuity handling and SER. 

This PR implements option (b), which re-evaluates two terms in an attached manner as illustrated above.

mworchel:
Hi all, 

I'd like to share my (and @ugogon's) thoughts on the technical details of this solution, following our exchange.

A potential issue is that viewing each 3-point neighborhood in _isolation_ is difficult to motivate when trying to derive this from the path space integral or rendering equation. 

I agree that one can take the path space integral and reparameterize individual integrals to be over solid angles, which makes the intersection point "slide over the surface". However, when taking the derivative, the choice of parameterization should be consistent for each path. For example, for a single path of depth 5, one can choose a `SMMMM` parameterization (`S` = (hemi)sphere, `M` = surface), or choose an alternating scheme `MSMSM` to break the recursion in the derivative of the spherical parameterization.

The potential issue with the current solution is that it parameterizes points inconsistently: x_{k+1} at the current intersection is treated as if it were sampled from the surface, but this x_{k+1} becomes x_{k} at the next intersection and is treated as if it were sampled from the hemisphere. I've slightly modified the minimal examples (from the referenced paper) such that _both_ planes move with the same parameter. Notice that the setting is free of discontinuities in the material form (`prb_threepoint` has no discontinuity handling). There's a visible bias in the proposed solution:
grafik

In other words, the importance `Wi` and the radiance `Li` also depend on `pi` in the formula given by @ziyi-zhang, and treating neighborhoods in isolation is only valid if the product rule would yield the same result. I would expect the above solution to work in a setting where the scene parameter moves two surfaces that are separated by one additional bounce because then the parameterization would be consistent (I've not yet verified this). 

EDIT: We wondered if a possible solution could be the average of two alternating parameterizations `MSMSM...` and `SMSMS...`, both of them being somewhat local.

Another thing I'm not 100% sure about is how the boundary terms of the different parameterizations relate to each other (the boundary terms of the material form and the spherical parameterization are different). Maybe this also has an effect in the experiment I've shown.

njroussel:
Hi @mworchel !

Ziyi brought me up to speed on some of the discussions you had.

As far as we can tell, the issue you pointed out was actually due to a small implementation issue on our end, rather than a flaw in the theory. In the first iteration of this PR, emitter sampling didn't correctly handle cases where the emitter is moving. This is only really an issue if the emitter is a textured area emitter, and therefore lead to small discrepancies in the third configuration you shared.

We test quite a few different setups that include moving geomtery but **do not have any discontinuities**. You can find the scenes in `test_ad_integrators.py`, the list is defined [here](https://github.com/mitsuba-renderer/mitsuba3/blob/6861dfc6a2f47e823b15005bbe3d592953b1ce98/src/integrators/tests/test_ad_integrators.py#L1041-L1049). We actually extended these and tested them with multiple moving shapes, and didn't find any issues.

In this whole process, we do believe we've also found some issues in the implementation of the boundary integral in the `*_projective` integrators. (Hence, why we found/built the aforementioned discontinuity-free scenes :sweat_smile:). That's an orthogonal topic that we'll look into; most likely as continuation of #1684.

We're comfortable with merging this PR as is now. Of course, if you find new issues/insights we're happy to continue the discussion in this thread (if I'm not mistaken, we should still be able to comment on a merged PR). Having the code merged, just makes is easier for us to work on. We're obviously still open to changing/fixing it :smile:

PR #1681, opened 2025-07-18
Fixes to the `blendbsdf` and `blendphase` plugin

bathal1:
The `blendbsdf` plugin can produce biased renderings under certain conditions:
- When sampling, we don't compute the correct pdf, which results in incorrect MIS weights
- If one of the nested BSDFs has a null component, the `volpath` plugin will miss some energy as the BSDF does not implement a `eval_null_transmission` method.

The tests run on this plugin seem to only use the same (opaque) BSDF twice, which may explain why we did not notice this before.

Below is some test scene containing 2 spheres and 2 planes:
- top sphere: `blendbsdf` mixing a `null` and `diffuse` BSDF
- bottom sphere: `mask` BSDF equivalent to the top one
- top plane: `blendbsdf` mixing 2 `diffuse` BSDFs
- bottom: `blendbsdf` mixing a `diffuse` and a `null` BSDF

Rendering with `volpath` on `master`:
pr_before

Rendering with `volpath` on this branch:
pr_after

Rendering the same scene in Cycles:
pr_blender

I also tested the MIS issue in a simpler scene where the expected pixel value can be determined analytically.

### DIFF
--- a/src/bsdfs/blendbsdf.cpp
+++ b/src/bsdfs/blendbsdf.cpp
@@ -134,6 +134,8 @@ public:
         if (dr::any_or<true>(m0)) {
             auto [bs0, result0] = m_nested_bsdf[0]->sample(
                 ctx, si, (sample1 - weight) / (1 - weight), sample2, m0);
+            Float pdf1 = m_nested_bsdf[1]->pdf(ctx, si, bs0.wo, m0);
+            bs0.pdf = weight * pdf1 + (1.f - weight) * bs0.pdf;
             dr::masked(bs, m0) = bs0;
             dr::masked(result, m0) = result0;
         }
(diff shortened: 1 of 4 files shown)


njroussel (commented):
Shouldn't we also be blending the evaluations? I think the tests were passing because we blend `0` and `1`.

Could you also please add a small test, or modify an existing one to check this behavior.

lnuic:
Hi @bathal1,

Thanks for pointing out this bug! I’ve updated the sampling weight calculation to properly account for sampling from a blended BSDF. A similar approach can be seen in Mitsuba 0.6 ([reference](https://github.com/mitsuba-renderer/mitsuba/blob/master/src/bsdfs/blendbsdf.cpp#L196-L205)).

I also found the same issue in `blendphase` and applied the fix there as well.

Could you please re-run the test scene you mentioned earlier to confirm if the results match Cycles?

njroussel (approved):

PR #1715, opened 2025-08-20
Fix minor parser issues

dvicini:
This PR fixes two issues I found in the new parser:

1) The test `test28_parameter_substitution_warnings` relied on the ordering of the `kwargs` being passed to a Python function. This ordering is not guaranteed and the test failed if a different order was obtained at runtime. I modified the test to use parameter keys of different lengths. The parser internally sorts by length, so this makes the test stable under arbitrary kwarg ordering

2) The `file_location` function should check if `state.nodes` is empty before trying to access. I was getting an invalid index issue otherwise.


wjakob:
Thanks Delio, these both look great to me!

PR #1717, opened 2025-08-21
Fix smooth face logic in blender shape

Speierers:
Simple fix in the Blender plugin for meshes that have smooth faces but no surface normals defined.


wjakob:
Thanks!

PR #1718, opened 2025-08-21
Add eval_3 to uniform texture plugin

Speierers:
This patch implements ` UniformSpectrum::eval_3()` which was missing.


wjakob:
Thanks!

PR #1721, opened 2025-08-22
Bugfix Sunsky

matttsss:
This PR addresses an issue where the decision whether the sun was hit was made twice for the same ray and yielding different results due to numerical errors. This caused fireflies that were most visible using the `ptracer` integrator, and can be seen on the garage door, the tower's roof and the balcony rail of the following scene:

### Before changes:
scene_before

### After changes:
scene_after


wjakob:
Thank you, Mattéo!

PR #1722, opened 2025-08-22
Extra check when updating `UniformSpectrum`

matttsss:
Currently, updating a `UniformSpectrum` with a `mi.ScalarColor3f` is valid and caused it to have a buffer of size 3 in `m_value`. This provokes a silent broadcast to a wavefront of 3 colors when calling `eval3`. 

A check is thus added to warn the user instead of causing an invalid size operation in JIT compilation.


wjakob:
Thanks. The CI is kind of flaky today..

PR #1723, opened 2025-08-28
Bump and normal map improvements

dvicini:
## Description
This pull request addresses two well-known issues in rendering of bump & normal maps:

1) Back facing normals: Shading normals may face away from the viewing vector, leading to black artifacts. This PR adds a simple flipping strategy to flip normals to be valid. This solves the most obvious visual artifacts.

2) Shadow falloff / terminator: Normal/bump maps may lead to overly harsh shadow termination. This PR implements https://link.springer.com/chapter/10.1007/978-1-4842-4427-2_12 to counteract this. Blender 4.5 implements the same method: https://projects.blender.org/blender/blender/pulls/135380

This PR modifies bump & normal map with those fixes. Both are on by default, but can optionally be disabled to revert back to a "text book" normal/bump mapping implementation without these production tricks. Credits to Philippe Weier for bringing my attention to those fixes and earlier experiments.

## Test renderings 
The following renderings show various test scenes. Left is current master, right is with the changes in this PR enabled.

**test_bumpmap_path**

image
image

**test_normalmap_path**

image
image

**test_normalmap_roughconductor_path**

image
image

**test_normalmap_identity_roughconductor_path** (as expected, flat normal maps should not change at all)

image
image


dvicini:
This is ready to be reviewed now. I updated all the render tests and added unit tests for normalmap settings.

njroussel:
I've just rebased and updated the `mitsuba-data` submodule (I've merged https://github.com/mitsuba-renderer/mitsuba-data/pull/21).

wjakob:
Hi @dvicini -- like in the other commit, could I ask you to document this change? (In this case, dropping a note to the two heuristics and linking the paper in the normal map docs would be perfect)

dvicini:
I added the paper reference and mention of the heuristic to both bump and normal map documentation, anything else that is missing?

wjakob:
Thanks you, this looks good. One more question from me: in the following correction, can you explain why the "facing the incident direction" holds after updating `n`? It was not immediately clear to me.

```diff
+        if (m_flip_invalid_normals) {
+            // Ensure that shading normals are always facing the incident direction.
+            Mask flip = Frame3f::cos_theta(si.wi) * dr::dot(si.wi, n) <= 0.0f;
+            n[flip] = Normal3f(-n.x(), -n.y(), n.z());
+        }
+```

dvicini:
The idea is to keep the "z" component unchanged, as for a valid normal map this should point upwards. We then flip the other two coordinates, which should correspond to this:

image

(figure from https://jo.dreggn.org/home/2017_normalmap.pdf)

I don't have a real proof of correctness though - so entirely possible I missed something. Experimentally, it does seem to work on our test scenes and produce "plausible" results.

PR #1732, opened 2025-09-02
Python plugins for Blender shader nodes

Speierers:
This PR provides Python texture plugins that should cover most of the Blender shader graph nodes.

Remaining TODOs:

- [ ] Proper testing
- [ ] Add-on code to instanciate those plugins when traversing the shader graph.


wjakob:
Hi Sébastien,

these look great, and it would be amazing to merge them into Mitsuba. But as you mentioned, there is no testing infrastructure at all, and that makes it tricky to accept your contribution as-is. Are you planning to work on these remaining TODOs?

Thank you!

PR #1733, opened 2025-09-02
Prevent NaN values for normals of triangles with zero area

dvicini:
The vertex normal computation in JIT mode should check for zero-area triangles. Such triangles might happen accidentally during an optimization and we do not want triangles with zero areas to lead to NaN normals for all adjacent vertices. 

This PR adds a simple check to prevent this. This actually mirrors the scalar code, which seems to already have this special case check (in that case, likely to work with imperfect existing assets)

### DIFF
--- a/src/render/mesh.cpp
+++ b/src/render/mesh.cpp
@@ -411,7 +411,10 @@ MI_VARIANT void Mesh<Float, Spectrum>::recompute_vertex_normals() {
 
         // --------------------- Kernel 2 starts here ---------------------
 
-        normals = dr::normalize(normals);
+        // Normalize and fall back to dummy value in case of zero-length vector.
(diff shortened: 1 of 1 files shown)


njroussel:
Hi @dvicini 

Isn't there still a risk that you get NaNs at a specific vertex if all the triangles it's a part of are zero-area? At the beginning of the second kernel we normalize the `normals` variable, which might still contain `(0, 0, 0)` if all faces were invalid. We should maybe also initialize it to `(1, 0, 0)` there, just in case. I'm sure there's a more robust way of handling this than setting this arbitrary value, but that will do for now. 

Feel free to merge the PR once you've made that small change :) and thanks for pointing this out

dvicini:
Yes that's true, I fixed this now too and will wait for CI and then squash and merge 👍

PR #1736, opened 2025-09-08
[WIP] Add Python bindings for field<T> types

dvicini:
This adds Python bindings for the `field<T>` types, such that we can expose the Sensor's `m_to_world` parameter. 

Work in progress


dvicini:
The basic version appears to be working, but there is one open question: `field.h` defines a specialized implementation for `TraversalCallback::put` for field objects. I am not sure what's the best way to also support that in Python. In C++, we can do `cb->put(m_to_world, ...)`, where `m_to_world` is a `field<T>` and will automatically unpack `T`.

Alternatively, we could ask for users to manually unpack fields in Python. But then it's unclear how to best raise an error in case this is forgotten? I.e., `cb.put(self.m_to_world, ...)` should raise an error, and `cb.put(self.m_to_world.value, ...)` should work)

I am open to suggestions or alternative solutions. The goal of the current PR was to solve the issue with the Sensor class, while keeping C++/Python APIs as close as possible (i.e., one could imagine a type of sensor where exposing `to_world` is undesirable, similarly how we do not expose `to_world` for all emitters either)

wjakob:
Hi Delio,

binding this kind of wrapper type seems somewhat complicated to me: you'd have to do it for every `T`, and those internal types then leak out into the Python API.

How about instead defining a nanobind type caster that strips away the `field<>` wrapper and accesses the underlying device-side value? 

You need to implement `from_python` and `to_python` functions that recursively dispatch to the default type caster. You could take a look at a basic one like `nanobind/stl/pair.h` to see how these generally work.

dvicini:
Yes that makes sense. I will try the type caster approach (I have some experience with those anyway)

dvicini:
Changed it to a type caster - what do you think? It seems to work.

(history is messy - please squash if you merge this)

wjakob:
LGTM, feel free to merge once the tests pass.

PR #1740, opened 2025-09-09
Allow multiple Python objects to refer to the same `Object*`

### LINKED ISSUE #1738: Duplicate python object binding
```py
import drjit as dr
import mitsuba as mi

mi.set_variant('cuda_ad_rgb')

scene = mi.load_dict(mi.cornell_box())
shapes = scene.shapes_dr()
print(shapes)
params = mi.traverse(scene)

```

This code throws because `set_self_py()` was called on an intrusive counter that had already transferred ownership to Python. But it's common that users modify the scene before traversing parameters?
> intrusive_counter::set_self_py(0x55f321112408): a Python object was already present![1]    3768768 IOT instruction (core dumped)

njroussel:
## Description

In many cases, Mitsuba will try to return a derived class rather than a "base plugin" class. For example, `scene.shapes()` will return `Mesh` objects instead of `Shape` ones when possible, or a `ProjectiveCamera` rather than a `Sensor` for `scene.sensors()`. However, some functions might still return (intentionally) return the base class. This can lead to situations where two python objects refer to the same C++ object.

This PR adjusts Mitsuba's use of nanobind's intrusive reference counting to 1) allow this situation to occur and 2) guarantee correct lifetimes of all objects. 

Fixes #1738

## Testing

Simple regression test.

### DIFF
--- a/src/render/tests/test_scene.py
+++ b/src/render/tests/test_scene.py
@@ -317,7 +317,7 @@ def test11_sample_silhouette_bijective(variants_all_ad_rgb):
     assert dr.allclose(valid_samples, valid_out, atol=1e-6)
 
 
-def test_enable_embree_robust_flag(variants_any_llvm):
+def test12_enable_embree_robust_flag(variants_any_llvm):
 
     # We intersect rays against two adjacent triangles. The rays hit exactly
     # the edge between two triangles, which Embree will not count as an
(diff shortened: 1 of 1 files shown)


njroussel:
Needed to add `nb::detail::nb_inst_set_state(po, true, false);` such that any new Python object will not runs its own destructor. We'd run into double-free situations if the new Python objects also ran destructor - we only want the first Python object that refers to the C++ object to trigger the destructor call.

wjakob:
Perfect! LGTM.

PR #1753, opened 2025-09-25
Fix `to_string` for Python-created meshes

dvicini:
The `to_string` method of meshes was not working for meshes that are created in Python using the properties only constructor. The reason for this appears to be that the C++ constructor that is referred to in the bindings was protected. Making it public makes it work. I added a simple unit test that fails before the fix and now passes.

This is a bug that confused quite a few new users, in particular when using mitsuba in colab/jupyter notebooks, where objects at the end of cells get printed by default.

### DIFF
--- a/include/mitsuba/render/mesh.h
+++ b/include/mitsuba/render/mesh.h
@@ -47,6 +47,9 @@ public:
          ScalarSize face_count, const Properties &props = Properties(),
          bool has_vertex_normals = false, bool has_vertex_texcoords = false);
 
+    // Creates an empty mesh.
+    Mesh(const Properties &props);
+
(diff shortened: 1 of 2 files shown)


njroussel (approved):
@dvicini  The change looks fine to me. I'm just curious how these two things are related. Where does nanobind require the constructor to be public? Or is it not even related to nanobind?

dvicini:
It does seem related to nanobind, but I do not quite understand the details. My theory is that with the protected constructor something goes wrong when initializing the trampoline structures in Python? 

I also couldn't see any other issue with the code, and indeed with the constructor made public it works.

PR #1756, opened 2025-09-26
Add assertion on align parameter of `tensor_io.write()`

matttsss:
This PR adds an assertion on the `align` parameter of the `tensor_io.write` function.
Previously, a user had no warning if he mistakenly passed a dictionary of tensors as the `align` parameter and the written file would just be empty.


njroussel (approved):
LGTM
That's indeed a bit of an easy mistake to do :sweat:

PR #1763, opened 2025-10-01
Add missing property binding

ziyi-zhang:
I want to load a scene (dict/xml) as node IR, modify it (add/remove shape), then load as mitsuba scene.
The following binding is missing for this.


wjakob:
LGTM!

PR #1767, opened 2025-10-03
Fix embree memory leak

dvicini:
This fixes a memory leak on the Embree acceleration data structure in non-LLVM (= scalar modes). In LLVM mode, we enqueue a host callback that eventually cleans up the Embree state. In scalar mode, this PR now adds the corresponding code to immediately free up Embree structures when the scene state is cleaned up.


wjakob:
Thank you!

PR #1659, opened 2025-07-06
``conftest.py`` refactored: metric collection, filtering/sorting, lower test memory usage

wjakob:
This PR bundles 3 changes that improve the Mitsuba test suite.

Change 1: Avoid bound types in `@pytest.parameterize`
========

Python objects passed to PyTest's `@pytest.parameterize` aren't reliably garbage collected when the interpreter shuts down. Why this happens would be worth tracking down at some point (this is a larger Python/GC issue and not specifically related to nanobind/Mitsuba/DrJit).

When such instances are bound by nanobind, they cause noisy leak warning messages to appear after the test suite terminates. The solution is simple: let's just not use nanobind types (e.g. `Transform4f`) in `@pytest.parameterize`. The commit adapts one test file, and the associated warnings go away following this change.

Change 2: `conftest.py` refactored: metric collection, filtering/sorting
========

This commit overhauls `conftest.py`, which provides infrastructure (fixtures and customization) for Mitsuba's PyTest-based test suite.

**1. Metric collection via** `--collect-metrics`

   New command-line parameter enables performance profiling of tests.
   Available metrics:

   - `memory`: Peak CPU/GPU memory usage
   - `time`: Wall-clock runtime
   - `launches`: Number of Dr.Jit kernel launches, cache statistics

   After test execution, a report displays the 100 most resource-intensive tests for each metric. Multiple metrics can be collected simultaneously using comma-separated values (e.g., `--collect-metrics=time,memory`). This helps identify tests that slow down CI or fail due to memory constraints on lower-end GPUs.

**2. Improved test filtering**
   The test suite previously generated many test cases that would ultimately be skipped due to unavailable or non-functional variants. This resulted in unnecessary overhead from creating and attempting to run invalid test configurations. The updated `conftest.py` now pre-filters these invalid configurations during test collection, preventing them from being generated in the first place. This reduces test discovery time and eliminates spurious skip reports in the test output.

**3. Variant-based test sorting**
   Tests parameterized by variant are now grouped to minimize variant switches. Previously, the test suite changed variants thousands of times during execution. Each variant change incurs overhead:

   - Python module reloading
   - `@dr.syntax` transformation
   - Python bytecode recompilation

   This optimization reduces the number of variant switches (which previously triggered warnings when `@dr.syntax` ran more than 1000 times) and improves overall test suite performance.

**4. Simplified variant fixture construction**
   Refactored code for creating variant fixtures to improve maintainability.

Change 3: Reduce memory usage of CI test suite
========

[...]

PR #1801, opened 2025-11-17
Skip unavailable tests instead of removing them

dvicini:
The latest `conftest.py` removes tests with unavailable variants, as explained in #1659. This makes sense, as the code previously used a somewhat inefficient way to skip tests with missing variants.

However, this seems to cause a minor issue with the latest pytest versions (>= 8.4.1). With these, test files without any valid tests now produce a non-zero exit code (exit code `5` indicating _No tests were collected_). This can for example happen when running `test_optixdenoiser` in a setting without a GPU, or `test_polarizer` when no polarization variant has been built.

This PR modifies the logic in `conftest.py` to mark tests for skipping, using the same (fast) logic as originally added in #1659. On my system, both with and without PR the test suite (LLVM only, `-m "not slow"`) takes around 50s. I do not observe a significant performance issue from having pytest report the skipped tests instead of removing them entirely.

It's a somewhat minor change, but makes it such that tests that do not fail consistently return exit code 0 :) Maybe worth double checking performance on other systems, if there are known challenges (e.g., on certain OS?)

### DIFF
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -306,10 +306,6 @@ def pytest_collection_modifyitems(config, items):
 
     variant_items = []
     for item in items:
-        # Skip tests with empty variant groups
-        if 'NOTSET' in item.name:
-            continue
-
         # Try to extract variant from test name
         match = pattern.search(item.name)
         if match:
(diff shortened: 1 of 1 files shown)


merlinND:
I don't know about performance implications, but personally I like having the tests marked as 'skipped' (with a clear reason) instead of having them disappear :+1:

wjakob:
In those few files, how about just adding a dummy test function? I hesitate to change the test discovery mechanism.

wjakob:
Just a quick question: we can't really know ahead of time which variants the user has. So there always needs to be a way of generating tests at run time. I have not reviewed the PR in detail but don't fully understand how this can be compatible with marking tests as skipped.

dvicini:
I only modified the `pytest_collection_modifyitems` function, which is called _after_ the tests have been collected. It appears that pytest allows to just mark tests as skipped after the initial collection (instead of dropping them). The mechanism to decide which tests can and cannot run is exactly identical to before.

dvicini:
Would it be possible to merge this? As far as I know, there isn't any (measurable) performance penalty from marking tests as "skip" instead of removing them completely.

PR #1812, opened 2025-12-04
Fix support for NumPy 2.4

dvicini:
NumPy 2.4 removes the `np.trapz` function in favor of `np.trapezoid`. However, NumPy < 2.0 does not have this function, so we need to conditionally pick which one to use for now (until we decide to drop support for NumPy < 2.0 and Python 3.8)


dvicini:
(Let's hold off with merging until we decide on whether to keep Python 3.8 or not)

wjakob:
Thanks!

PR #1814, opened 2025-12-08
Update docs to require Python 3.9 instead of 3.8

dvicini:
This PR updates the minimum Python version to 3.9 instead of 3.8. Since wheels are only built for Python >= 3.9, this is really just a change in documentation at the moment. In a later step, we can possibly simplify code since will no longer support 3.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ for details on this.
 
 ### Requirements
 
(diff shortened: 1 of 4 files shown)


wjakob:
Good catch! Thanks, Delio.

PR #1838, opened 2026-01-27
Port Ellipsoids bbox computation to JIT

Speierers:
This PR ports the `Ellipsoids::recompute_bbox()` to using the JIT for the LLVM and CUDA variants, significantly speeding up the scene instantiation and update when working with large assets.


njroussel:
Thanks ! :100:

PR #1868, opened 2026-04-02
Bump drjit and improve typechecking stubs

WeiPhil:
Hi,
These changes aim at _improving_ typechecking in Mitsuba, in particular for pytype, but also partially for mypy and Pyrefly. Most of the changes are in the `stubs.pat` file to patch the existing stub generation to improve basic type compatibility between DrJit/Mitsuba/NumPy, and support various initialization styles without false positives from the typechecker.

Disclaimer: these changes are just a first step towards better typechecking, but they should already provide a good basis for further improvements. More importantly, some of the changes here are required for us (internally) to continue testing and improving the typechecking features, so I hope this gets merged even if it is not yet a completely flawless solution.

I'll be happy to continue improving on this as I encounter new typechecking issues, etc.
Best,
Philippe


wjakob:
Thanks you, these look great. It may be impossible to get _perfect_ type checking given the limitations of Python typing. But any step in that direction is welcome.

PR #1869, opened 2026-04-04
Animated transforms & motion blur

dvicini:
## Description
This PR adds motion blur to Mitsuba to better support realistic light transport simulation in dynamic environments. It enables the rendering of moving scenes and sensors (e.g., a camera on a moving robot).

Specifically, this PR implements the following

- **AnimatedTransform**: This adds an `AnimatedTransform` class. This class stores a list of key frames and linearly interpolates using standard decomposed transform interpolation. This is similar to the version that existed in Mitsuba 2, see [link](https://github.com/mitsuba-renderer/mitsuba2/blob/62863cb3b58ab81ac5161908f39c2bfbf928cb9d/include/mitsuba/core/transform.h). The new implementation follows Mitsuba 3's conventions, and includes fast paths for common cases. It carefully manages host and device state to allow users to update transforms after loading (e.g., to animate a static sensor loaded from an XML scene).
- **Time-varying sensors & emitters**: We replace the `Endpoint`'s world transform with the animated version. All sensors and emitters have been updated to allow using the animated transform. This is fairly non-intrusive, but will change how the user accesses the `to_world` transform using the traversal mechanism (effectively, adding one extra level of nesting, since `AnimatedTransform` is a traversable class, and not a Dr.Jit struct).
- **Instance-level animation**: For shapes, we support animation via the instancing transform. While custom shapes (e.g. sphere) could also directly support animation, using instancing reduces the overall clutter and code changes. Most importantly, mesh geometry anyway requires the instancing mechanism to be rendered with motion blur, and does not offer a more "direct" path. We use OptiX' and Embree's decomposed transform interpolation to prevent artifacts, and avoid naive direct matrix interpolation.

All changes are done in way that they should not introduce significant overheads for static scenes.

## Testing

The following tests were added
- Comprehensive unit tests for the `AnimatedTransform` class.
- A few specific animation tests for some of the most important cases. In particular, we test animated instancing in a few different configurations.
- Render tests for both moving / deforming shapes as well as moving lights

## Limitations

- Shape animation requires evenly spaced keyframes. This is a requirement imposed by OptiX & Embree. We check for this and raise an error if the user provides invalid keyframes. We do not provide extensive tooling to define animation paths, the assumption is that any complex animation would be generated outside of Mitsuba and passed as "baked" keyframes for rendering.
- We for now only support affine transformations (e.g., no shearing). This seems sufficient for object-level animation. 
- We did consider differentiability or reconstruction under motion blur, this would need to be explored separately in the future.

## Future work
In future PRs, I would like to add the following features to further improve realism & generality:
- **Deformation motion blur:** Support for keyframed mesh animation rendering. This can likely be done by introducing a new `animated_mesh` plugin that uses a list of nested meshes.
- **Rolling shutter:** Implementing rolling shutter for sensors would be fairly straightforward, and can be important for some usecases (most real digital cameras use a rolling shutter).

## Checklist


include/mitsuba/core/animated_transform.h
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Create an empty animated transformation
+    AnimatedTransform() = default;
+
+    /// Initialize from a constant transformation
+    AnimatedTransform(const ScalarAffineTransform4f &trafo);
+
+    /// Initialize from a map of keyframes
+    AnimatedTransform(
+        const std::map<ScalarFloat, ScalarAffineTransform4f> &keyframes);
```
wjakob, line 50: std::map is a very complex data structure. I don't think it is a good choice to pass data. Better to use std::vector<std::pair<..>>

include/mitsuba/core/animated_transform.h
```diff
@@ -0,0 +1,131 @@
...
+    /// Equality comparison operator
+    bool operator==(const AnimatedTransform &other) const;
+
+    /// Inequality comparison operator
+    bool operator!=(const AnimatedTransform &other) const {
+        return !operator==(other);
+    }
+
+    /// Returns the host-allocated keyframes of the animated transform.
+    const std::map<ScalarFloat, Keyframe> &keyframes() const {
```
wjakob, line 78: ditto

include/mitsuba/render/optix/shapes.h
```diff
@@ -31,12 +32,16 @@ struct MiOptixAccelData {
     HandleData linear_curves;
     HandleData custom_shapes;
 
+    std::vector<void*> motion_transforms;
```
wjakob, line 35: Why void*?

src/core/animated_transform.cpp
```diff
@@ -0,0 +1,365 @@
...
+    const auto &kf1      = it1->second;
+    ScalarMatrix3f s     = dr::diag(dr::lerp(kf0.S, kf1.S, t));
+    ScalarQuaternion4f q = dr::slerp(kf0.Q, kf1.Q, t);
+    ScalarVector3f tr    = dr::lerp(kf0.T, kf1.T, t);
+    return ScalarAffineTransform4f(
+        dr::transform_compose<ScalarMatrix4f>(s, q, tr),
+        dr::transpose(dr::transform_compose_inverse<ScalarMatrix4f>(s, q, tr)));
+}
+
+MI_VARIANT void AnimatedTransform<Float, Spectrum>::parameters_changed(
```
wjakob, line 129: The decision to have multiple DynamicBuffers looks like a driver complexity. My suggestion would be that you pack everything into a single DynamicBuffer with stride 12 (time, scale, quaternion, position, 1 unused entry). That will then also enable 3xvec4 /vectorized loads to access these elements.

This is also the direction I want to go for the Mesh class. (Pack positions, normals, etc. in one buffer). The vectorized loads/stores/atomics are much faster.

wjakob: dr::slice is in general an extremely expensive function to use. At least 1 full PCI express transaction per call in CUDA mode. In LLVM mode it requires synchronizing with nanothread, inserting a memcpy, waiting for it to finish, etc.

wjakob (commented):
Hi Delio -- this looks great. Some initial feedback on a few things that stuck out to me. I am also curious (discussed separately on slack) whether the non-animated version keeps on working in scalar mode (including Mitsuba's builtin kdtree without Embree) and if the indirection to copy a transform to the stack of the caller (instead of direct loads that can be inlined) has an impact on performance.

include/mitsuba/core/animated_transform.h
dvicini: Makes sense, switched to std::vector<std::pair<...>>.

include/mitsuba/core/animated_transform.h
dvicini: Changed it to vector of pairs

include/mitsuba/render/optix/shapes.h
dvicini: Changed it to specify the actual type, `OptixSRTMotionTransform*`

src/core/animated_transform.cpp
dvicini: I think the packed layout makes sense, in terms of performance. Just to makes sure: In that case, the `traverse` mechanism would expose the packed buffers and it would be on the users to do the unpacking & packing in their code, in case they would like read/modify the animated transform?

It wouldn't affect the default case of non-animated transforms, where we can expose the underlying "plain" transform just as before.

dvicini: I've updated the code now, the internal changes were easy to do.

dvicini:
Hi Wenzel,

The non-Embree version appears to still work. I addressed the other feedback. The only open point is performance evaluation. I tried on the `living-room` scene and the initial tests didn't show a measurable performance difference in scalar mode. I can try a few more scenes before we merge this, the code generally needs another careful look by myself to be considered mergeable.

KBentley57:
@dvicini 

This work has a lot of promising features for our use case.  Is there anything I can do to help?  We'd create motion blur by simply rendering many times, which is largely inefficient.

dvicini:
Hi Kyle,
Thanks! It's close to being ready in a first version, no specific help needed at the moment.

wjakob:
Hi Delio, do you see any way to avoid the parameters `.to_world` change? That's a pretty disruptive change that could potentially be handled in a more compatible way.

docs/src/key_topics/animation.rst
```diff
@@ -0,0 +1,433 @@
...
+      - Default
+      - Description
+    * - ``shutter_open``
+      - ``0.0``
+      - Time at which the shutter opens
+    * - ``shutter_close``
+      - ``0.0``
+      - Time at which the shutter closes
+
+Each camera ray then draws its time uniformly from that interval, using one
```
wjakob: This seems too detailed for the user-facing documentation. (Going into the MC sampling code)

docs/src/key_topics/animation.rst
```diff
@@ -0,0 +1,433 @@
+.. _sec-animation:
+
+Animation and motion blur
```
wjakob: This whole section feels somehow a bit "chatty". I think that some details can be safely removed.

docs/src/key_topics/animation.rst
```diff
@@ -0,0 +1,433 @@
...
+    * - Parameter
+      - Shape
+      - Contents
+    * - ``times``
+      - ``(N,)``
+      - Keyframe times, strictly increasing
+    * - ``scale``
+      - ``(N, 3)``
+      - Per-axis scale factors
+    * - ``shear``
```
wjakob: Do we actually need to support interpolatable shear? This sounds like it would make the decoding of the animation frames quite a bit more expensive.

include/mitsuba/render/accel_embree.h
```diff
@@ -86,6 +86,22 @@ struct EmbreeAccel {
...
+    ScalarFloat time_min = 0.f;
+    ScalarFloat time_max = 0.f;
+
+    /**
+     * Map a ray time onto the [0, 1] range that Embree expects
+     *
+     * Times outside of the scene-wide keyframe range clamp to the first/last
+     * keyframe. The range is defined by *all* animated instances.
+     */
+    template <typename Value> Value normalize_ray_time(const Value &time) const {
```
wjakob: Is this a no-op when the scene does not make use of animations?

PR #1872, opened 2026-04-15
Further refinements for typechecking

WeiPhil:
Hi,
Another few refinements for better typechecking. 
In this pass we have:
- A few more handy core types initialisation (TensorXf, FloatStorage, etc..), including some extra numpy compat. 
- A small change in `mesh.h` to use mitsuba rater than drjit types (which allows the typechecker to use our patched core types automatically).
- ScalarRay variants that don't have bindings, yet are generated by many Transforms automatically (I don't think these will ever be useful, but it seems cleaner to add these "dummy" bindings than trying to patch all Transforms in the stubs.pat file)
- And last but not least, updated scores that ensure LLVM and CUDA typechecking always prevail on the scalar mode (which in some combinations of modes wasn't always ensured)

With these changes we have reached a (surprising) high percentage of correct typechecking in our codebase already!

Best,
Philippe

### DIFF
--- a/include/mitsuba/render/mesh.h
+++ b/include/mitsuba/render/mesh.h
@@ -114,7 +114,7 @@ public:
     template <typename Index>
     MI_INLINE auto face_indices(Index index,
                                 dr::mask_t<Index> active = true) const {
-        using Result = dr::Array<dr::uint32_array_t<Index>, 3>;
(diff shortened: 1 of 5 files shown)


wjakob:
Thank you, Philippe!

PR #1874, opened 2026-04-29
[WIP] Add support for Metal backend and ray-tracing

Speierers:
# Mitsuba: Metal (Apple Silicon GPU) variants

This PR adds **Metal variants** to Mitsuba, enabling GPU-accelerated
rendering on Apple Silicon (M-series, macOS 13+) without going through
CUDA / OptiX. After this PR, the `metal_rgb`, `metal_ad_rgb`,
`metal_spectral`, etc. variants are available alongside the existing
`scalar_*`, `llvm_*`, and `cuda_*` variants.

This PR depends on the companion **drjit-core** and **drjit** PRs that
introduce the underlying Metal JIT backend.

## Summary

- New Metal variants for every existing Mitsuba spectrum × polarisation
  combination, except `*_double` (Apple GPUs have no FP64).
- New `accel_*_metal` / `ray_*_metal` Scene paths, parallel to the
  existing OptiX paths.
- All shape plugins gain optional Metal hooks; meshes, spheres, disks,
  cylinders, ellipsoids, sdfgrids, linear / B-spline curves,
  shapegroups, and instances all render correctly under Metal variants.
- On Apple platforms, the default variant set switches to include
  `metal_ad_rgb` instead of `cuda_ad_rgb` so a fresh checkout builds
  out of the box.
- Every Metal-specific path is gated on `MI_ENABLE_METAL`, so non-Apple
  builds — and Apple builds that do not opt in — are unaffected.

## How rendering works on Metal

The Metal acceleration structure path (in the new `src/render/scene_metal.inl`)
mirrors `scene_optix.inl`:

1. **BLAS construction.** Each shape contributes one or more entries:
   * Triangle meshes use `MTLAccelerationStructureTriangleGeometryDescriptor`
     and reuse Drjit-managed vertex / index buffers (looked up via
     `jit_metal_lookup_buffer`).
   * Linear and B-spline curves use Apple's native
     `MTLAccelerationStructureCurveGeometryDescriptor`.
   * Spheres, disks, cylinders, ellipsoids, and SDF grids use a
     `MTLAccelerationStructureBoundingBoxGeometryDescriptor` with
     custom MSL intersection programs.
2. **TLAS construction.** A `MTL::InstanceAccelerationStructure` is
   built over the BLAS list, with one
   `MTLAccelerationStructureUserIDInstanceDescriptor` per instance.
3. **Intersection function table.** A single `MTL::Library` is compiled
   from the MSL source embedded in
   `include/mitsuba/render/metal/intersection_functions.h` and is
   bundled with the per-shape data buffers into an IFT.
4. **Scene registration.** The TLAS, the resource list, the IFT, and
   the geometry-types mask are handed to drjit-core via
   `jit_metal_configure_scene`, which returns a JIT-variable handle.
5. **Ray tracing.** `ray_intersect_*_metal` lowers a ray-trace call to
   `jit_metal_ray_trace`, threading the per-scene handle so that the
   compiled kernel binds the correct TLAS / IFT for each trace
   operation.

### New files

| File | Purpose |
|------|---------|
| `src/render/scene_metal.inl` (~890 LoC) | Counterpart of `scene_optix.inl`. BLAS / TLAS construction, scene registration, `ray_intersect_*_metal`, JIT traversal callbacks. |
| `include/mitsuba/render/metal/intersection_functions.h` (~570 LoC) | MSL source (as a C++ raw string literal) for the sphere / disk / cylinder / ellipsoids / sdfgrid intersection programs. Math is a 1:1 port of the OptiX programs in `src/shapes/optix/*.cuh`. |
| `include/mitsuba/render/metal/shapes.h` | POD layouts (`SphereData`, `DiskData`, `CylinderData`, `EllipsoidData`, ...) shared between host C++ and the MSL source. `static_assert`s lock down the byte sizes. |

### Shape API

`include/mitsuba/render/shape.h` gains a small set of virtual methods
(all guarded by `#if defined(MI_ENABLE_METAL)`) that non-mesh shapes
implement to plug into the Metal acceleration-structure builder:

```cpp
virtual size_t   metal_aabb_count() const { return 1; }
virtual size_t   metal_primitive_data_size() const { return 0; }
virtual size_t   metal_total_data_size() const;
virtual uint32_t metal_intersection_function_index() const { return 0; }
virtual void     metal_fill_aabb_data(void *out) const;
virtual void     metal_fill_primitive_data(void *out) const { }

// Native curves
virtual int      metal_curve_kind() const { return 0; }
virtual void     metal_get_curve_data(uint32_t *cp, size_t *n_cp,
                                       uint32_t *idx, size_t *n_seg) const { }

// Instancing
virtual std::vector<ref<Shape>> metal_instance_children() const { return {}; }
virtual void     metal_instance_to_world(float out[12]) const;
```

This is deliberately analogous to the existing OptiX hooks
(`optix_geometry_flags`, `optix_build_input`, `optix_fill_hitgroup_records`,
`optix_prepare_geometry`) on the same class.

Concrete plugins implement these:

* **sphere, disk, cylinder** — single-AABB BLAS, per-primitive data block
  consumed by a custom MSL intersection function.
* **ellipsoids** — multi-AABB BLAS, one `EllipsoidData` per ellipsoid.
* **sdfgrid** — multi-voxel BLAS with a custom buffer layout (header +
  voxel indices + sample grid).
* **linearcurve, bsplinecurve** — native HW curve geometry descriptor.
* **instance** — exposes the shapegroup's children + the per-instance
  `to_world` transform; the scene builder constructs one BLAS per child
  and a TLAS instance descriptor that points to it.

### Per-IFT-slot combined buffer layout

Worth flagging for review: Metal's intersection function table only
has one binding per `[[buffer(N)]]` slot, so two shapes of the same type
cannot each own a distinct buffer at the same slot. We work around this
by **concatenating** the per-primitive data of every shape of a given
type into one combined buffer per type, and using
`MTLAccelerationStructureUserIDInstanceDescriptor::userID` as a per-
instance offset into that buffer. The MSL function reads
`[[user_instance_id]] + primitive_id` to pick the right region. This is
documented at the top of `intersection_functions.h`.

SDFGrid is the exception (data is shape-level, not per-primitive) and
keeps a per-shape buffer; multiple SDFGrids in one scene currently emit
a warning.

### Scene plumbing

`include/mitsuba/render/scene.h` adds the symmetric `accel_*_metal`,
`ray_*_metal`, `static_accel_*_metal`, and `traverse_1_cb_*_metal`
declarations. `src/render/scene.cpp` dispatches to them via
`if constexpr (dr::is_metal_v<Float>)` next to the existing CUDA branch.
Stub implementations for non-Metal builds raise `NotImplementedError`.

### Variants

`resources/mitsuba.conf.template` defines the new variants:

```
metal_mono                metal_ad_mono
metal_mono_polarized      metal_ad_mono_polarized
metal_rgb                 metal_ad_rgb
metal_rgb_polarized       metal_ad_rgb_polarized
metal_spectral            metal_ad_spectral
metal_spectral_polarized  metal_ad_spectral_polarized
```

No `*_double` variants — Metal has no FP64.

On Apple platforms, the default variant set in `CMakeLists.txt` becomes:

```
scalar_rgb,scalar_spectral,metal_ad_rgb,llvm_ad_rgb,llvm_ad_spectral
```

(other platforms remain unchanged, with `cuda_ad_rgb` in place of the
Metal variant).

### Other shared-code touchpoints

Small "extend the CUDA branch to also cover Metal" edits:

* `include/mitsuba/core/{fwd.h,spectrum.h}` and `src/core/spectrum.cpp` —
  CIE colour-space tables get a Metal instance;
  `color_management_static_initialization` takes an extra `bool metal`
  argument.
* `include/mitsuba/core/distr_2d.h` — host-pinned upload path now also
  triggers for Metal.
* `src/core/python/drjit_v.cpp`, `src/python/{main,main_v,alias}.cpp` —
  surface `metal` as a backend name and expose `MI_ENABLE_METAL`.
* `src/python/python/chi2.py` — performs the chi² accumulation in numpy
  float64 when running on Metal (since the GPU has no FP64).
* `src/integrators/tests/test_ad_integrators.py` — coerce `TensorXf` to
  Python `float` for failure-print formatting.
* `src/conftest.py` — register `metal_*` variants and `any_metal` /
  `vec_backends_once_*` fixture combinations.

## New dependencies (Apple only)

All gated on `MI_ENABLE_METAL` — non-Apple builds and opt-out Apple
builds link no Metal libraries.

* **System frameworks** linked into `mitsuba-render`:
  * `Metal`
  * `Foundation`
* **`metal-cpp` headers**, added as a private include path on
  `mitsuba-render`. The headers come from
  `ext/drjit/ext/drjit-core/ext/metal-cpp` so the dependency is shared
  with drjit-core and not duplicated in this repository.

No new first- or third-party libraries are added.

## Build

`MI_ENABLE_METAL` is auto-derived from the configured variants:

```cmake
if (MI_VARIANTS MATCHES "metal_")
  set(MI_ENABLE_METAL ON)
endif()
```

i.e. it turns on as soon as any `metal_*` variant is enabled. To build
explicitly:

```bash
# pick variants in mitsuba.conf, then:
cmake -B build -S .
cmake --build build
```

## Limitations

1. **No FP64 on Metal.** Variants that need double precision must use
   LLVM or CUDA.
2. **SDFGrid multi-shape.** Only one SDFGrid per scene currently
   routes correctly; multiple SDFGrids emit a warning until per-primitive
   shape pointers are wired up.
3. **`accel_parameters_changed_metal` rebuilds from scratch** rather
   than refitting the existing BLAS / TLAS. Matches the original
   `accel_init` path but is suboptimal for shapes whose vertex data is
   changing every iteration; refit support is a future optimisation.
4. **Curves** use the native HW curve descriptor, restricting us to the
   basis functions Apple supports (linear, cubic B-spline). Other curve
   types would require a custom-intersection fallback.
5. **Metal does not support zero-copy GPU↔CPU sharing**, so workflows
   that round-trip data through PyTorch or NumPy incur an explicit copy.

## Test plan

- [ ] All existing variant tests pass on Linux / Windows (no regression
      for non-Apple users).
- [ ] All existing variant tests pass on macOS without any `metal_*`
      variants enabled (`MI_ENABLE_METAL=OFF`).
- [ ] On macOS with `metal_ad_rgb` enabled:
  - [ ] `pytest src/` passes for the Metal variant (plugin tests:
        BSDFs, phase functions, sensors, films, samplers, integrators,
        and shapes — meshes, spheres, disks, cylinders, ellipsoids,
        sdfgrids, linear / B-spline curves, instances).
  - [ ] AD integrator tests (`test_ad_integrators.py`) match reference
        gradients within tolerance.
  - [ ] Cornell-box renderings agree with the LLVM / CUDA variants
        (within FP32 noise).
  - [ ] Curves, ellipsoids, SDFGrid, and instance scenes render
        correctly (`render_cornell_*.py` scripts at the repo root).
- [ ] `mi.MI_ENABLE_METAL` reflects the build configuration; available
      variants list reports the `metal_*` set when the backend is
      present.


wjakob:
Closed in favor of #1897.

wjakob:
Thanks for your PR, Sébastien. In contrast to the Dr.Jit part where I merged your PR followed by a sequence of refactors, I found that the Mitsuba part needs a more fundamental rewrite of the entire ray tracing layer that is done in the commit linked above.

PR #1876, opened 2026-05-03
MergeShape: improve mesh merge

lnuic:
The merge plugin previously reduced N input meshes per group via N successive acc.merge(next) calls. Each Mesh::merge runs dr::concat over the full accumulated buffers, giving O(M·N) total work per group, where M is the combined mesh size. This shows up clearly when many small meshes share few BSDFs.

Replace the linear fold with a tree-balanced pairwise reduction: O(M·log N) per group. 
On a 27k-mesh / 65-BSDF test case: load time ~1670 s → ~16 s.
We might want to implement a batched version of mesh merge ?


wjakob:
Hi Lovro -- thank you very much for this. I rolled in those changes as part of commit 9f4c644691ecba0a0c5ea4b288b6205fdfe3b4c3.

PR #1884, opened 2026-05-18
`ImageBlock`: Use packet scatter for per-pixel accumulation and remove kahan summation

lnuic:
Updates `ImageBlock` to use the packet scatter instead of N separate
per-channel scatter-reduces. On the drjit-core side, the CUDA
backend now runs a single warp-reduction shared across
all channels of the packet (previously the reduction ran once per channel).

This gives ~10% speedup on average. On Blackwell-class GPUs
(CC ≥ 120 with a recent CUDA driver 13.2+), the same path additionally
uses vector atomic reductions, for another ~10% on top.

Since the speedup comes entirely from faster atomic aggregation, 
scenes with short per-ray work benefit most.

## Benchmarks

Configurations: butterfly scope (rows) is how contributions are pre-aggregated within a warp, and atomic form (columns) is how the leader stores them.

|                            | Scalar atomics | Vector atomics |
| ---                        | ---            | ---            |
| No butterfly               | `per_channel`  | `v4_only`      |
| Per-channel butterfly      | `baseline`     | n/a            |
| Single packet butterfly    | `bfly_packet`  | `bfly_v4`      |

Tested on the gallery scenes.

**1–8 spp**

bench_results_summary_smaller_spp

**16–1024 spp:**

bench_results_summary

### DIFF
--- a/include/mitsuba/render/imageblock.h
+++ b/include/mitsuba/render/imageblock.h
@@ -105,14 +105,6 @@ public:
      *   are random and will in any case be subject to thread divergence (e.g.
      *   in a particle tracer that makes random connections to the sensor).
      *
-     * \param compensate
-     *   If set to \c true, the implementation internally switches to
-     *   Kahan-style error-compensated floating point accumulation. This is
-     *   useful when accumulating many samples into a single precision image
-     *   block. Note that this is currently only supported in JIT modes, and
-     *   that it can make the accumulation quite a bit more expensive. The
-     *   default is \c false.
-     *
      * \param warn_negative
      *    If set to \c true, \ref put() will warn when writing samples with
      *    negative components. This test is only enabled in scalar variants by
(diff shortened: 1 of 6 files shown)


wjakob:
Could I ask you to add a second commit to this PR that removes the Kahan summation feature from ImageBlock? (and any plugins that set this value). This feature turned out to be completely pointless in retrospect since it did not address precision issues with very long runs (e.g. 1M SPP).

lnuic:
Done, added a commit that drops Kahan summation from `ImageBlock` and from `hdrfilm`/`specfilm`. 
Also added benchmark plots to the PR description.

wjakob:
This is awesome. I also tested it on the Metal backend. Despite lacking packet atomic hardware support, rendering the material preview test scene there goes from ~1 sec to ~850msec.

PR #1898, opened 2026-06-19
Re-enable parallel scene parsing

dvicini:
The current parser tries to synchronously instantiate objects without dependencies. However, due to the way the scene is traversed (depth first), this just leads to the scene loading being serialized.

Without this change, a scene with 500 unique 1kx1k PNG textures takes ~30s to load. With this change, it takes 2s.

### DIFF
--- a/src/core/parser.cpp
+++ b/src/core/parser.cpp
@@ -1802,6 +1802,15 @@ std::vector<ref<Object>> instantiate(const ParserConfig &config, ParserState &st
     if (state.empty())
         Throw("No nodes to instantiate");
 
+#if defined(MI_ENABLE_LLVM) || defined(MI_ENABLE_CUDA) || defined(MI_ENABLE_METAL)
+    // Flush pending side effects here and now, to avoid potentially dirty
+    // user-provided Dr.Jit arrays/tensor from being propagated to plugin
+    // loaders running on a different thread. Side effects are queued in
+    // per-thread data structures.
+    if (config.parallel && !string::starts_with(config.variant, "scalar_"))
+        jit_eval();
+#endif
+
     std::vector<Scratch> scratch(state.size());
     instantiate_node(config, state, scratch, 0);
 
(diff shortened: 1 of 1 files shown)


wjakob:
Any idea about the failing CI tests? Is that pre-existing or some behavior surfaced by the additional parallelism?

dvicini:
There appears to indeed be a real failure. The following test fails:

```python
def test04_load_tensor(variants_vec_backends_once):
    grid = dr.full(mi.TensorXf, 0.9, [3, 3, 3])
    grid[:, 0, :] = 0.0
    # dr.eval(grid) <--- Inserting this fixes it, pre-evaluating `grid`
    texture = mi.load_dict({
        'type': 'volume',
        'volume': {'type': 'gridvolume', 'data': grid},
    })
    si = dr.zeros(mi.SurfaceInteraction3f)
    si.p = mi.Point3f([0.4, 0.4, 0.3], [0.5, 0.0, -0.5], [0.5, 0.4, 0.5])
    assert dr.allclose(dr.mean(texture.eval(si)), [0.9, 0.0, 0.0])
    assert dr.allclose(texture.max(), 0.9)
```

I think there is some issue due to the side effects of the scatter operation being unevaluated before the `load_dict` call. I think the brute-force fix is to insert ` nb::module_::import_("drjit").attr("eval")(dict);` into `load_dict`'s Python bindings. However, I feel like this might just be hiding a deeper Dr.Jit issues? Not sure what's the expected behavior here.

wjakob:
Can you include this?
```
@@ -1804,6 +1804,15 @@ std::vector<ref<Object>> instantiate(const ParserConfig &config, ParserState &st
     if (state.empty())
         Throw("No nodes to instantiate");

+#if defined(MI_ENABLE_LLVM) || defined(MI_ENABLE_CUDA) || defined(MI_ENABLE_METAL)
+    // Flush pending side effects here and now, to avoid potentially dirty
+    // user-provided Dr.Jit arrays/tensor from being propagated to plugin
+    // loaders running on a different thread. Side effects are queued in
+    // per-thread data structures.
+    if (config.parallel && !string::starts_with(config.variant, "scalar_"))
+        jit_eval();
+#endif
```

dvicini:
Thanks, that fixes it! I will wait until the CI passes to squash & merge

wjakob:
Awesome, all green. Thank you for catching this issue!

PR #1906, opened 2026-06-24
Fix B-spline curve silhouette gradients

fknfilewalker:
This PR fixes bugs related to the `bsplinecurve` shape plugin in connection with the projective integrator.

* Fixes the mismatch between texture UVs and curve parameterization UVs for the `bsplinecurve` shape
  that caused the projective integrator to produce wrong gradients.
* Rebuild the position/normal partials from the consistent local frame and
  its analytic derivatives instead of the Frenet torsion, which depends on
  the discontinuous third derivative and injected spurious gradients at the
  segment borders.

### Debug Views
<table>
  <tr>
    <th width="33%">Fixed</th>
    <th width="33%">Metal (Broken)</th>
    <th width="33%">LLVM (Broken)</th>
  </tr>
  <tr>
    <td>
      grad_comparison_fixed
    </td>
    <td>
      grad_comparison_metal_broken
    </td>
    <td>
      grad_comparison_llvm_broken
    </td>
  </tr>
</table>


wjakob:
cc @ziyi-zhang

fknfilewalker:
Still has some issues at the curve openings.

Issue #1911, opened 2026-06-26
3.9.0: metal_ad_rgb does not render analytic sphere/cylinder shape (works on llvm_ad_rgb)


ArnoCo:
The analytic `sphere` / `cylinder` shapes are invisible under the `metal_ad_rgb` backend. The identical scene renders the sphere correctly under `llvm_ad_rgb`. Only the variant differs. Same behaviour for the cylinder shape.

Versions:
- mitsuba==3.9.0
- drjit==1.4.0
- Python 3.11.7
- macOS 15.6.1, Apple M2, Macbook Air

Minimal reproducer (manually requires swapping the backends):
```python
import mitsuba as mi
import matplotlib.pyplot as plt

mi.set_variant("metal_ad_rgb")

scene_dict = mi.cornell_box()
scene_dict["sphere"] = {
    "type": "sphere",
    "radius": 0.45,
    "center": [0, -0.55, 0.1],
    "bsdf": {
        "type": "diffuse",
        "reflectance": {"type": "rgb", "value": [0.2, 0.25, 0.7]},
    },
}
scene = mi.load_dict(scene_dict)

img = mi.render(scene)

plt.imshow((img ** (1 / 2.2)).numpy().clip(0, 1))
plt.show()
```

ArnoCo:
@zhestyatsky I checked out your fix, and this seems to resolve the issue on my end! Thanks!

PR #1912, opened 2026-06-26
Fix rendering analytic shapes on metal

zhestyatsky:
Addresses issue https://github.com/mitsuba-renderer/mitsuba3/issues/1911 by removing an extra tag from all shapes


wjakob:
Thanks!

PR #1920, opened 2026-07-17
Force libstdc++.so.6 DT_NEEDED to fix symbol interposition segfaults

tjueterb:
## Description

Edit: The culprit was `gcc 16.1.1+r346+g4e03491b401d-3`, see comment below.

~On newer Linux toolchains (observed on Arch Linux with binutils 2.46.1, glibc 2.43, GCC 16 / Clang 22), binutils' default `--as-needed` drops `libstdc++.so.6` from `DT_NEEDED` when a library only references libstdc++ via vague-linkage (inline template) symbols — in my build, no produced `.so` file has a `libstdc++.so.6` dependency. `std::` symbols then bind to incomplete copies exported by vendored libraries (e.g. `libpugixml.so`) and crash, e.g. when printing any plugin. Because plugins are dlopen'd with `RTLD_DEEPBIND`, this is not fixable via `LD_PRELOAD`.~

~Fix: Force the `DT_NEEDED` entry at link time via `-Wl,--push-state,--no-as-needed -l:libstdc++.so.6 -Wl,--pop-state` in the top-level `CMakeLists.txt` (the `-l:` soname form is required; plain `-lstdc++` is dropped even under `--no-as-needed`). Gated behind `check_linker_flag()` and a Linux/GCC/Clang guard, so unaffected platforms and toolchains see no change. Set before `add_subdirectory(ext)`, so vendored libraries inherit the flags without changes under `ext/`.~

## Testing

`pytest src/bsdfs/tests/test_diffuse.py` segfaults before the fix and passes after.

## Checklist

- [ ] I have made corresponding changes to the documentation (not necessary)
- [ ] I have added tests that prove my fix is effective or that my feature works (not necessary)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,6 +298,21 @@ endif()
 
 set(CMAKE_INSTALL_RPATH "${MI_ORIGIN};${MI_ORIGIN}/../drjit")
 
+# Force a libstdc++.so.6 DT_NEEDED entry so std:: symbols never bind to copies in vendored libraries
+if (UNIX AND NOT APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)"
+    AND NOT CMAKE_CXX_FLAGS MATCHES "libc\\+\\+")
+  include(CheckLinkerFlag OPTIONAL RESULT_VARIABLE MI_CHECK_LINKER_FLAG)
+  if (MI_CHECK_LINKER_FLAG)
+    set(MI_LINK_LIBSTDCXX "-Wl,--push-state,--no-as-needed -l:libstdc++.so.6 -Wl,--pop-state")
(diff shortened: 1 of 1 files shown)


wjakob:
It seems wrong to hardcode `libstdc++.so.6` in the build. I am not doubting that there is a problem, but this doesn't seem like a robust long-term fix.

I have personally never experienced this issue, so it would first be helpful to know how to reproduce it.

tjueterb:
Turns out this was not a Mitsuba bug but a broken gcc version that was shipped on Arch.
`gcc 16.1.1+r346+g4e03491b401d-3` was the culprit, the bug has since been fixed in version `-4`.

The `gcc` package was missing the `libstdc++.so` symlink next to `libstdc++.a` in its own lib dir. Nothing to do with `--as-needed` as I originally wrote.

I am closing this as no change is needed.

Issue #1917, opened 2026-07-15
Large overhead for AOV rendering


dvicini:
I've encountered this today and thought I would leave a note here, even though I don't have a good solution suggestion at the moment:

It appears that AOV rendering currently comes with a pretty significant overhead. Even just outputting depth in addition to color can potentially more than double (!) the render time for simplistic scenes. Here are some statistics I collected using the metal backend on the Cornell box:

```
=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'box')
=========================================================

0 AOVs (Pure Path)                         | Avg: 444.99 ms
1 AOV (Depth)                              | Avg: 940.80 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 988.79 ms

=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'gaussian')
=========================================================

0 AOVs (Pure Path)                         | Avg: 479.03 ms
1 AOV (Depth)                              | Avg: 1205.28 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 1522.61 ms
=========================================================
```

For reference, here is the benchmark script I used:

```python
import argparse
import copy
import time
import mitsuba as mi
import drjit as dr

def main():
  parser = argparse.ArgumentParser(description="Raw Mitsuba 3 Cornell Box AOV Benchmark.")
  parser.add_argument("--spp", type=int, default=64)
  parser.add_argument("--width", type=int, default=1024)
  parser.add_argument("--height", type=int, default=1024)
  parser.add_argument("--num_frames", type=int, default=10)
  args = parser.parse_args()

  mi.set_variant("metal_ad_rgb")

  # Load base Cornell Box scene dictionary
  cbox_dict = mi.util.cornell_box()
  cbox_dict["sensor"]["film"]["width"] = args.width
  cbox_dict["sensor"]["film"]["height"] = args.height
  cbox_dict["sensor"]["sampler"]["sample_count"] = args.spp

  # Test configurations across filter types and AOVs
  filters = ["box", "gaussian"]

  for filter_name in filters:
    print(f"\n=========================================================")
    print(f" Cornell Box Benchmark ({args.width}x{args.height}, {args.spp} SPP, Filter: '{filter_name}')")
    print(f"=========================================================\n")

    # Construct clean scene dict for this filter run
    scene_dict = mi.util.cornell_box()
[...]

PR #1921, opened 2026-07-18
Performance: Improve AOV rendering, prevent duplicate rendering passes

dvicini:
Attempt to address issue #1917: Currently, the AOV integrator queries the nested RGBA integrator values both in it's `sample` function, as well as (separately) by calling `render` recursively in the `render` function.

Originally, the AOV integrator did not override `render`, so the `sample` function was correct. However, later, `render` was modified to query nested integrators that may override `render` themselves. That change should have removed the nested integrator query in `sample`, which is most likely an oversight.

**Update 21.7.2026:** The PR now correctly prevents duplicate rendering, while carefully updating the film's content. I also added a unit test to check the correct handling of nested integrators. The change required adding a new API to retrieve a film's storage, and a new API to write to image blocks. This PR also removes expensive per-sample heap allocations in scalar mode.

With the change, I get these timings on an RTX A5000:

```
=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'box')
=========================================================

0 AOVs (Pure Path)                         | Avg: 129.42 ms
1 AOV (Depth)                              | Avg: 138.58 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 144.48 ms

=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'gaussian')
=========================================================

0 AOVs (Pure Path)                         | Avg: 345.43 ms
1 AOV (Depth)                              | Avg: 371.60 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 517.41 ms
```

Without the change, I get:
```
=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'box')
=========================================================

0 AOVs (Pure Path)                         | Avg: 130.62 ms
1 AOV (Depth)                              | Avg: 316.08 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 414.57 ms

=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'gaussian')
=========================================================

0 AOVs (Pure Path)                         | Avg: 352.10 ms
1 AOV (Depth)                              | Avg: 1857.04 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 3638.11 ms
=========================================================
```

### DIFF
--- a/src/integrators/aov.cpp
+++ b/src/integrators/aov.cpp
@@ -85,7 +85,7 @@ the material, this value might only be an approximation.
  */
 
 template <typename Float, typename Spectrum>
-class AOVIntegrator final : public SamplingIntegrator<Float, Spectrum> {
+class AOVIntegratorImpl final : public SamplingIntegrator<Float, Spectrum> {
 public:
     MI_IMPORT_BASE(SamplingIntegrator)
     MI_IMPORT_TYPES(Scene, Shape, Sensor, Sampler, Medium, BSDFPtr, ShapePtr)
@@ -103,36 +103,10 @@ public:
         dUVdy,
         PrimIndex,
         ShapeIndex,
-        IntegratorRGBA
     };
 
-    AOVIntegrator(const Properties &props) : Base(props),
-        m_integrator_aovs_count(0) {
-        std::vector<std::string> tokens = string::tokenize(props.get<std::string_view>("aovs"));
-
-        // First pass: collect integrators and their RGBA channels
-        std::vector<std::pair<std::string, Base*>> integrators_with_names;
-        for (auto &prop : props.objects()) {
-            Base *integrator = prop.try_get<Base>();
-            if (!integrator)
-                Throw("Child objects must be of type 'SamplingIntegrator'!");
-            std::string name(prop.name());
-            integrators_with_names.push_back({name, integrator});
-            m_integrators.push_back(integrator);
-            m_aov_types.push_back(AOVType::IntegratorRGBA);
-            m_aov_names.push_back(name + ".R");
-            m_aov_names.push_back(name + ".G");
-            m_aov_names.push_back(name + ".B");
-            m_aov_names.push_back(name + ".A");
-            m_integrator_aovs_count += 4;
-        }
-
-        // Second pass: collect AOVs from integrators
-        for (auto &[name, integrator] : integrators_with_names) {
-            std::vector<std::string> aovs = integrator->aov_names();
-            for (auto aov_name: aovs)
-                m_aov_names.push_back(name + "." + aov_name);
-        }
+    AOVIntegratorImpl(std::string_view aovs_spec) : Base(Properties()) {
+        std::vector<std::string> tokens = string::tokenize(aovs_spec);
 
         for (const std::string &token: tokens) {
             std::vector<std::string> item = string::tokenize(token, ":");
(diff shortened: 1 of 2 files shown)


dvicini:
not quite working yet - seems a bit more complex than I thought, since we need both the return value (with develop = True) and the film's internal buffer to store the correct values...

dvicini:
Revised the code and it now correctly updates the film, so even with `develop = False`, we should get the right content.

The updated timings are nearly identical as with the original simplistic fix:
```
=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'box')
=========================================================

0 AOVs (Pure Path)                         | Avg: 132.46 ms
1 AOV (Depth)                              | Avg: 142.73 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 149.20 ms

=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'gaussian')
=========================================================

0 AOVs (Pure Path)                         | Avg: 355.77 ms
1 AOV (Depth)                              | Avg: 383.39 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 533.74 ms
=========================================================
```

wjakob:
Hi Delio,

this is a great catch!

I think there is a problem with the channel arithmetic in render(): ``channel_count() - m_aov_names.size()`` gives the number of non-AOV channels, which is 4 for pixel_format="rgb" but 5 for "rgba", while each nested integrator only reserves 4 slots (RGBA). It seems like you increased `NB_OVERRIDE()` by one too much.

Your modified version of `sample()` fills the nested integrator slots with zeros, but the ImageBlock splats them through the reconstruction filter anyway (Packet atomic scatter-adds with a partial set of zeros don't benefit from the usual Dr.Jit optimizations). This probably accounts for some of the remaining lost performance.

This could be avoided by rendering the integrator's own AOVs through a small internal SamplingIntegrator whose `aov_names()` only contains the `aovs="..."` channels (skipping that pass when the list is empty), snapshotting each pass with `film->develop(raw=true),`assembling the final image as one tensor with the existing channel-slice helpers in `aov.cpp`, and then preparing the film once with the full channel list and installing that tensor via `film->clear()` + `film->put_block(...)`. `put_block()` adopts the tensor by reference when the target is cleared and the sizes match, so this costs nothing extra. It would also make Film::storage() and `ImageBlock::copy_channels_from()` unnecessary, since everything goes through existing API.

dvicini:
Thanks Wenzel! This all makes sense, and using the films develop(raw=True) feature allows to get rid of the inelegant new APIs. It's still quite a bit of logic to deal with edge cases (e.g., no nested integrators, or no actual AOVs and only nested integrators), but at least it's contained in this one plugin.

I've updated the code, and the timings are indeed a bit better:

```
=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'box')
=========================================================

0 AOVs (Pure Path)                         | Avg: 130.86 ms
1 AOV (Depth)                              | Avg: 139.96 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 141.23 ms

=========================================================
 Cornell Box Benchmark (1024x1024, 64 SPP, Filter: 'gaussian')
=========================================================

0 AOVs (Pure Path)                         | Avg: 349.32 ms
1 AOV (Depth)                              | Avg: 375.81 ms
5 AOVs (Depth, Normal, Albedo, Pos, UV)    | Avg: 439.95 ms
=========================================================
```

dvicini:
@wjakob I cleaned this up some more (removed unused variable, simplified copy channels logic) and it should now be good to merge.

wjakob:
Approved, thanks!

PR #1922, opened 2026-07-22
Performance: Introduce a MergeInstance shape

dvicini:
This PR introduces a `MergeInstance` shape. By automatically merging matching instance plugins during scene parsing, we can significantly reduce JIT tracing cost for scenes containing many (>100s) instances. Instead of tracing each instance separately, we just create a single plugin and locally fetch the appropriate transforms. In practice, this can reduces JIT tracing cost from impractical (>30s for 1000 objects) to completely practical (<0.1s).

This is work in progress and needs more polish to be reviewable. It may also make sense to discuss if we want to keep the ability to have individual instance objects (when `optimize=False` during scene loading) or not. Another option would also be to fold the `MergeInstance` plugin into the regular instance plugin, by simply allowing that plugin to take a number of transforms.

### DIFF
--- a/include/mitsuba/core/parser.h
+++ b/include/mitsuba/core/parser.h
@@ -86,6 +86,9 @@ struct ParserConfig {
     /// Merge compatible meshes (same material) into single larger mesh
     bool merge_meshes = true;
 
+    /// Merge instances of the same ShapeGroup into MergeInstance shapes
+    bool merge_instances = true;
+
     /// Constructor that takes variant name
     ParserConfig(std::string_view variant) : variant(variant) {}
 };
--- a/include/mitsuba/render/accel_embree.h
+++ b/include/mitsuba/render/accel_embree.h
@@ -48,7 +48,8 @@ struct EmbreeAccel {
 
     // --- Declarative traversal ---
     DRJIT_TRAVERSE(EmbreeAccel, accel_handle, func_handle,
-                   occlude_handle, shapes_registry_ids)
+                   occlude_handle, shapes_registry_ids,
+                   batch_element_ids)
 
     /// Native Embree scene, lifetime tied to ``accel_handle`` in JIT variants.
     RTCSceneTy *accel = nullptr;
@@ -65,6 +66,20 @@ struct EmbreeAccel {
     UInt64 func_handle;
     UInt64 occlude_handle;
     DynamicBuffer<UInt32> shapes_registry_ids;
+
+    /// For LLVM path: maps geom_id → batch element index within a
+    /// MergeInstance.  Value is (uint32_t)-1 for non-batch shapes.
+    DynamicBuffer<UInt32> batch_element_ids;
+
+    /// MergeInstance.  Value is (uint32_t)-1 for non-batch shapes.
+    ///
+    /// For MergeInstance shapes that emit N Embree geometries, all N geom_ids
+    /// map to the same \c m_shapes index.
+    std::vector<uint32_t> geom_id_to_shape_idx;
+
+    /// Maps Embree geom_id → batch element index within a MergeInstance.
+    /// For non-batch shapes, the value is (uint32_t)-1.
+    std::vector<uint32_t> geom_id_to_batch_idx;
 };
 
 NAMESPACE_END(mitsuba)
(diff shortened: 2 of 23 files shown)


frettini:
Hello @dvicini, 

I'm very interested in this PR, as we have found that scenes with many instances can incur large memory overhead (see discussion https://github.com/mitsuba-renderer/mitsuba3/discussions/1794). Since then, I have done [further tests](https://github.com/eradiate/mitsuba3/blob/instance_list/src/eradiate_plugins/shapes/instancelist.cpp) exploring an approach similar to the `MergeInstance` you propose i.e. not rely on the expansion mechanism, which unsurprisingly significantly reduced the memory usage. Our use case is probably very different from yours, we are mostly using Mitsuba with a scalar variant, with a KDTree, in double precision.

With this in mind I have a couple of suggestions/questions: 
- It would be very useful if the plugin could also accept transforms as an Nx4x4 tensor. 
- Are there any plans to extend support to KDTrees? If not I'd be happy to look into it, do you think it would be feasible?

Many thanks!
Nicolae

dvicini:
Hi Nicolae,

Interesting, I didn't see this discussion before. What I implemented here focuses on the JIT performance, trying to reduce the number of distinct plugins in the scene. The problem I am trying to solve is that currently the JIT compilation very quickly becomes the bottlenecks for high instance counts.

To your questions:
1) I am not sure, I haven't looked into that. I guess in python dict it's fairly easy to pass a tensor, but I am not sure if this would then also work in the XML format. Maybe a mismatch here is acceptable, since some other plugins already support special cases in Python (e.g., directly passing an `mi.Bitmap` instance as texture).

2) I only had a preliminary look and it seemed like it would be a bit of work. If you have suggestions on how to do this while minimizing the overall changes, let me know.

frettini:
Regarding tensors, an approach similar to the one in `grid.cpp` could be considered, where a separate parameter accepts a to_world tensor. I will have a look at kdtrees and how it can be adapted.

frettini:
Hello @dvicini , 
It seems possible to integrate `MergeInstance` to kdtrees, it would require some changes to `scene_native.inl`, the `kdtree`, and a new entry point to ray intersections in `Shape`:
- In `scene_native.inl`, the `shape_registry_ids` and `batch_element_ids` would be populated to have a shape per batched instance. 
- In the kdtree, a sort of `shape_to_instance_map` would be added to retrieve the instance index from a given shape index.
- In both the kdtree and scene_native, some logic to properly handle the shape_index, prim_index, and instance_index (the tricky part, especially in llvm mode).
- New overloads to `ray_intersect_preliminary_scalar` and `ray_test_scalar` that accept a `prim_index` in addition to the usual `ray` parameter. For most shapes, this would default to the non-indexed version except for `MergeInstance` which would use the index to gather the correct transform and delegate to `ShapeGroup`. 

Let me know if you want to go ahead with this, and what would be your preferred way to integrate those changes (PR into your branch, cherry picking a separate branch, pushing directly on the branch).

Thanks, 
Nicolae

wjakob:
@frettini @dvicini I plan to remove the very outdated KD-tree (`scene_native.inl`) in an upcoming release and replace it with TinyBVH (which supports double precision ray tracing). So it's probably not worth it for Delio to implement a fix. It would be good to coordinate with the Eradiate team on how to make this transition smooth. Maybe it would allow you to use stock Mitsuba without having to ship a custom build.

frettini:
@wjakob , a move to TinyBVH would be very welcome! In any case, MergeInstance remains an important feature for us. 
We have diverged quite a bit by implementing local majorant/minorant structures and further variance reduction methods so we'll definitely need further coordination and integration to use stock Mitsuba. We can get back to you once @leroyvn is back from holidays.

wjakob:
Are you mainly working with `scalar` mitsuba or the LLVM version? If the former, why? (Let's maybe continue the discussion on Slack since this is not related to this ticket).

wjakob:
Thank you Delio. After some consideration, I will close this PR in favor of #1945. I benchmarked both approaches. Doing things at the scene level (the #1945 approach) was consistently faster (~1.7-1.8x render time, 4x trace time) when rendering a large Blender scene that makes heavy use of instancing.

wjakob:
(Is it OK to delete the batch_instancing branch?)

dvicini:
Thanks! Yes please go ahead and delete this branch.

PR #1926, opened 2026-07-29
Fix AOV splatting in the AD integrator forward/backward passes

fknfilewalker:
An AOV should not be scaled by the sensor importance. Furthermore, multiplying a `Float` AOV by the `Spectrum` sensor weight promotes it to a `Color3f`, which then sits in the channel list handed to `block.put`, and `put` only accepts `Floats`, which results in a `TypeError` aka crash.

Fix:

Do not multiply `δaov` with `weight`.

## Checklist


wjakob:
Thanks!

PR #1928, opened 2026-07-31
Fixes for Bsplinecurve

fknfilewalker:
Fixes some bugs which caused problems with the Bsplinecurve. 

* Clamp the segment index in partials(). 
* Use no caps when using metal, similar to all other backends. 
* Make the segment count opaque, in order to not run into out of bounds issues when the curve count changed when using function freezing.

### DIFF
--- a/src/render/metal_accel.mm
+++ b/src/render/metal_accel.mm
@@ -433,7 +433,9 @@ build_impl(const std::vector<BlasEntry> &blases,
                                                           ? MTLCurveBasisBSpline
                                                           : MTLCurveBasisLinear;
                         gd.curveType                = MTLCurveTypeRound;
-                        gd.curveEndCaps             = MTLCurveEndCapsSphere;
+                        gd.curveEndCaps             = bspline
+                                                          ? MTLCurveEndCapsNone
+                                                          : MTLCurveEndCapsSphere;
                         gd.opaque                   = YES;
                         [geoms addObject: gd];
 
--- a/src/shapes/bsplinecurve.cpp
+++ b/src/shapes/bsplinecurve.cpp
@@ -277,6 +277,7 @@ public:
         m_indices = dr::load<UInt32Storage>(indices.get(), segment_count);
         m_curves_prim_idx = dr::load<UInt32Storage>(curves_1st_prim_idx.get(),
                                                     curve_1st_idx.size() + 1);
+        m_segment_count = dr::opaque<Float>((ScalarFloat) segment_count);
 
         std::unique_ptr<InputFloat[]> positions =
             std::make_unique<InputFloat[]>(m_control_point_count * 3);
@@ -345,6 +346,7 @@ public:
             recompute_bbox();
             mark_dirty();
         }
+        m_segment_count = dr::opaque<Float>((ScalarFloat) dr::width(m_indices));
         Base::parameters_changed();
     }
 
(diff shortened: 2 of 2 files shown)


fallenmi (changes requested):
Requesting changes for one correctness issue.

The new `control_points` path should reject non-finite positions and radii before `recompute_bbox()`, matching the existing filename loader. On exact head `ee54703b8cc3b94b4e82ad08e291c013b505ed3a`, both `scalar_rgb` and `llvm_ad_rgb` accept `mi.TensorXf` inputs containing `NaN` or `+/-Inf`. Positive infinity produces non-finite bounds; scalar `NaN` produces an invalid bbox, while the JIT path can silently calculate bounds that omit the affected control point. The equivalent file inputs are rejected by the existing `invalid position data` / `invalid radius data` checks. Letting these values reach the shape therefore corrupts or truncates the bounds used by acceleration structures.

Please validate all four values of every in-memory control point before the transform/bbox work and add scalar plus JIT regression cases for tensor and flat-buffer inputs.

Verification: I compiled the exact-head `bsplinecurve.cpp` plugin with its exact bbox/interaction headers against the official Mitsuba 3.9.1 wheel runtime. The PR's focused tests 22-27 pass in all 18 available RGB variant cases, and a separate non-finite oracle reproduces the mismatch in `scalar_rgb` and `llvm_ad_rgb`. The independently generated current-`master` merge tree retains the same constructor and bbox code paths.

Disclosure: I used Codex to inspect the change, build the exact-head test plugin, and run the focused tests and regression oracle. I verified the reported behavior and the live PR state before preparing this review.

fallenmi (changes requested):
Requesting changes for one out-of-bounds regression in the new opaque segment-count path.

`m_seg_count_f` stores the integral segment count in a 32-bit `Float`, but `eval_parameterization()` and `partials()` then derive their index upper bound as `UInt32(segment_count) - 1u`. Counts above `2^24` are not all exactly representable. With 16,777,219 segments, the exact conversion produces `count_f == 16,777,220`; at `v == 1`, the new clamp returns segment index 16,777,219 even though the last valid index is 16,777,218. `cubic_interpolation()` then gathers `m_indices` out of bounds. With 16,777,217 segments, the rounded-down bound instead skips the valid final segment. The base implementation's host-integer upper bound remains exact in both cases.

Please retain an exact opaque integer count for index bounds/clamps, alongside the opaque `Float` used for normalization if needed. Please also add a focused freeze regression that replays one recording with curves having different segment counts; the existing freeze test reuses one curve and only changes control points, so it does not cover the behavior this patch fixes.

I verified this against exact head `11fe1d281bc34ca95996700ad5510c2edfb9356e` and the current live-master merge tree. A lightweight LLVM Dr.Jit oracle of the exact conversions reproduces both the one-past index and rounded-down last-segment cases. All four exact-head TeamCity contexts are green; the PR changes no tests.

Disclosure: I used OpenAI Codex to inspect the source and repository policy, reconstruct the current merge, run the focused arithmetic oracle, and prepare this review. I verified the exact revisions, reported behavior, interactions, CI, and conclusion before submission.

wjakob:
Hi @fknfilewalker,

these changes look good to me. Please let me know if/when they should be merged. 

I am not sure why our guest working on "sovereign AI", "defense-tech" and "algorithmic trading" has opinions about a PR on elliptic curves. My guess it's some kind of OpenClaw agent. Probably best to ignore.

fknfilewalker:
Great, should be good to go!

> I am not sure why our guest...

Really curious what the purpose of it is, though. Maybe it's something noble like improving worldwide FOSS code quality.

PR #1933, opened 2026-08-05
Major rewrite of the ``Mesh`` class

wjakob:
# Major rewrite of the `Mesh` class

This commit redesigns the `Mesh` class to address several flaws:

1. Mitsuba was not able to represent geometry that was geometrically continuous while having discontinuous normals or UVs.  Mesh loaders achieved this by cutting seams into the mesh, which caused normal creases, discontinuities in a differentiable renderer, and tearing under optimization.

2. Mitsuba computed piecewise constant tangents *per face*, causing renderings with anisotropic materials to take on a faceted appearance.

3. Mesh data was split over many separate buffers, which is bad for performance. The `compute_surface_interaction()` function (which is a key bottleneck) had to first load pointers to these buffers and then perform a large (~28) sequence of scalar gathers.

4. In the `SceneParameters` view, meshes were (awkwardly) exposed as flat/raveled arrays.

5. Mitsuba implemented its own scheme for tangent interpolation and shading coordinates, which was incompatible with the [MikkTSpace](https://github.com/mmikk/MikkTSpace) convention that is used by most professional digital content creation tools. This meant that normal maps authored in other tools did not look correct in Mitsuba.

6. It was not possible to assign different materials to the individual faces of a mesh without splitting it.

7. Mitsuba shipped a custom constructor to extract a corner-indexed mesh by bit-peeking into internal Blender data structures, which was fragile and often broke when new Blender versions changed the layout of these data structures.

This commit replaces the mesh representation, exposed parameters, construction API, and storage internals to address all of these flaws.

⚠️**WARNING**⚠️: This is an **API-breaking change**. Existing code that constructs meshes or touches mesh scene parameters will need to be updated. See the porting guide at the end of this message.

The Mesh class now exposes parameters as non-flat tensors. It smoothly interpolates tangents according to MikkTSpace. The representation is still an indexed triangle mesh, but with additional indirections (more on this below) that allow it to represent discontinuities affecting only a subset of parameters. The implementation internally packs data so that information characterizing individual faces or vertices can be obtained with more efficient packet load operations. On newer GPUs (NVIDIA Blackwell), the prelude in `compute_surface_interaction()` reduces to just 4 such instructions.

The new implementation also adds low-level support for per-face material IDs, though fully enabling this as a feature in Mitsuba will require additional follow-up work that is not part of this commit.

The Mesh class exposes the following fields, which are row-major tensors, with `F` faces, `V` vertices, `P` surface points, and `N` normal groups:

| Field | Shape | Contents |
| --- | --- | --- |
| `faces` | `(F, 3)` | vertex indices in `[0, V)` |
| `bsdf_index` | `[F]` | per-face material index |
| `position_index` | `[V]` | indices in `[0, P)` |
| `normal_index` | `[V]` | indices in `[0, N)` |
| `positions` | `(P, 3)` | |
| `normals` | `(N, 3)` | |
| `texcoords` | `(V, 2)` | |

Positions and shading normals are potentially *coarser* than the vertices, with `P <= N <= V`. The `position_index` and `normal_index` maps record which vertex belongs to which entry, so that copies split at a seam still reference the same underlying position or normal.

Reading the texture coordinate of corner `c` of face `f` is therefore a direct lookup,

```python
vertex_idx = faces[f][c]
texcoord   = texcoords[vertex_idx]
```

while positions and normals take one further step through their map, e.g:

```python
vertex_idx   = faces[f][c]
position_idx = position_index[vertex_idx] if len(position_index) > 0 else vertex_idx
position     = positions[position_idx]
```

An empty index map stands for the identity.

Rendering consumes a packed 8-float (32 byte) record per vertex. `pack()` (re)builds this packed representation when the user changes the tensors listed above.

Construction is a two-step process: a constructor that creates an empty mesh, followed by one of

- `from_fields()`, which accepts the field tensors listed above,
- `from_packed()`, which accepts the packed representation verbatim,
- `from_corners()`, which ingests corner-indexed meshes (e.g. from OBJ files or DCC applications like Blender) into that representation. The Blender-specific constructor and its `blender` shape plugin were deleted in favor of it.

A `Mesh(name, faces, positions, ...)` convenience constructor covers the common case in a single call.

## Porting guide

**Scene parameters**. The parameters exposed by `mi.traverse()` changed name and shape:

| Before (flat buffer) | After (tensor) |
| --- | --- |
| `vertex_positions` `[3V]` | `positions`, `(P, 3)` |
| `vertex_normals` `[3V]` | `normals`, `(N, 3)` |
| `vertex_texcoords` `[2V]` | `texcoords`, `(V, 2)` |
| `faces` `[3F]` | `faces`, `(F, 3)` |
| vertex attribute `[dV]` | `(V, d)`, and `(F, d)` per face |

The `position_index` `[V]`, `normal_index` `[V]`, and `bsdf_index` `[F]` entries are new and writable like the rest.

**Editing parameters**. Because the values are shaped tensors, the flat-buffer bookkeeping that used to surround an edit tends to disappear:

```python
# Before
v = dr.unravel(mi.Point3f, params['teapot.vertex_positions'])
v.z += 0.5
params['teapot.vertex_positions'] = dr.ravel(v)

# After
params['teapot.positions'] += mi.TensorXf([[0, 0, 0.5]])
```

An operation that does need the structure-of-arrays layout, such as applying a `Transform4f`, converts in either direction via `flip_axes`:

```python
# Before
v = dr.unravel(mi.Point3f, params['bunny.vertex_positions'])
params['bunny.vertex_positions'] = dr.ravel(trafo @ v)

# After
v = mi.Point3f(params['bunny.positions'], flip_axes=True)
params['bunny.positions'] = mi.TensorXf(trafo @ v, flip_axes=True)
```

Resizing/remeshing is supported. Simply write new tensors and calls `update()`, which validates the whole batch of changes and rebuilds the acceleration data strutures.

```python
params['shape.faces']     = mi.TensorXu(new_faces)
params['shape.positions'] = mi.TensorXf(new_positions)
params.update()
```

**Construction**. Building a mesh is now a single call that supplies the data up front, instead of allocating an empty mesh and filling its buffers through `mi.traverse()`:

```python
# Before
mesh = mi.Mesh("wavydisk", vertex_count=N, face_count=N - 1,
               has_vertex_normals=False, has_vertex_texcoords=False)
params = mi.traverse(mesh)
params['vertex_positions'] = dr.ravel(vertex_pos)    # mi.Point3f
params['faces'] = dr.ravel(face_indices)             # mi.Vector3u
params.update()

# After
mesh = mi.Mesh("wavydisk",
               faces=face_indices,                   # (F, 3) TensorXu
               positions=vertex_pos)                 # (P, 3) TensorXf
```

The constructor optionally also accepts `normals`, `texcoords`, `position_index`, `normal_index`, `bsdf_index`, `face_normals`, and `flip_normals`.

**Parameter handling and winding order**: The face winding order now strictly defines the orientation of the surface. The geometric normal of a face follows from the right hand rule applied to its positions. Shading normals must be compatible with this by being located in the same hemisphere. ``Mesh`` maintains this invariant, which means that it might need to change the winding order when a mesh is transformed by a `to_world` transformation. Meshes now bake `flip_normals` into the representation at construction/load time instead of handling this parameter when computing surface intersections.

**Method renames**. The remaining interface changes are mechanical:

| Before | After |
| --- | --- |
| `mi.Mesh(name, V, F, props, has_vertex_normals, has_vertex_texcoords)` | `mi.Mesh(name)` plus a construction call, or `mi.Mesh(name, faces, positions, ...)` |
| `faces_buffer()` | `faces()` |
| `vertex_positions_buffer()` | `positions()` |
| `vertex_normals_buffer()` | `normals()` |
| `vertex_texcoords_buffer()` | `texcoords()` |
| `attribute_buffer(name)` | `attribute(name)` |
| `has_vertex_texcoords()` | `has_texcoords()` |
| `add_attribute(name, dim, values)` | `add_attribute(name, tensor)` |
| `mesh.merge(other)` | `mi.Mesh.merge(shapes)` |


wjakob:
For posterity: a screenshot of the before/after rendering the default material test ball with per-face (**left**, old) and interpolated per-vertex tangents (**right**, new).

mp_before_after

PR #1937, opened 2026-08-13
Add eval scope guard to build_pmf in mesh.cpp

dvicini:
`build_pmf` may be invoked during symbolic vcalls (e.g., `sample_position` calls it) and it therefore also needs the new eval scope guard.


wjakob:
Thanks!

PR #1939, opened 2026-08-19
Add prbvolpath_sm: new unbiased differentiable volume integrator with sample matching

AuroraRyan0301:
This PR adds `prbvolpath_sm`, a volumetric integrator that estimates extinction gradients with sample matching.

> Ruihan Yu, Yu-Chen Wang, Jingwang Ling, Feng Xu, Shuang Zhao.
> **Sample Matching for Joint Extinction Gradient Estimation in Differentiable
> Volume Rendering.** ACM Transactions on Graphics 45(4), 2026.
> [paper](https://auroraryan0301.github.io/samplematching/static/papers/sample_matching_paper.pdf)
> · [supplementary](https://auroraryan0301.github.io/samplematching/static/papers/sample_matching_supplementary.pdf)
> · [project page](https://auroraryan0301.github.io/samplematching/)
> · [doi:10.1145/3811329](https://doi.org/10.1145/3811329)

Denser media scatter more light toward the camera and also block more of the light passing through. These two terms of the extinction gradient have opposite signs. `prbvolpath` estimates them at unrelated points; `prbvolpath_sm` estimates both at the same points on each path segment, which cancels most of the variance. The primal image is the same as `prbvolpath`; all changes are in the adjoint pass.

The adjoint pass runs as two kernels. The path replay loop writes one record per segment, and a second kernel reads the records and traces new rays for the gradient samples. `segment_slots` sets how many records a path may keep. Any number of slots is unbiased. By default it is `max_depth + 1`, so every segment gets a record. With one slot, a path keeps a single segment chosen by weighted reservoir sampling; this setting is also registered as `prbvolpath_sm_linear`, and the default as `prbvolpath_sm_quad`.

Spectral variants and extinction that varies across channels are supported. Polarized variants and forward mode are not, same as `prbvolpath`.

## Testing

Bunny cloud, one A40, 8000 iterations. PSNR is measured on the held-out view at 256 spp:

| | `prbvolpath` | `prbvolpath_sm_linear` | `prbvolpath_sm_quad` (default) |
|---|---|---|---|
| gradient variance, same spp | 1 | 0.29 | 0.25 |
| PSNR | 26.66 dB | 35.71 dB | 38.35 dB |
| time | 41.6 h | 22.6 h | 26.3 h |

extinction gradient mean and variance

Gradients agree with finite differences within 1% in RGB and spectral variants. The test suite passes on `cuda_ad_rgb`, `cuda_ad_spectral`, `llvm_ad_rgb` and `llvm_ad_spectral`.

## Checklist


AuroraRyan0301:
The two Linux builds are failing on this branch and the failure is mine, not
infrastructure. Posting what I know, because I cannot reproduce it and would
rather not guess at a fix.

**What happens.** `test06_gradients_colored_extinction[cuda_ad_rgb-config0]`
aborts the interpreter. Every test up to it passes (420 passed, then exit
code 134):

```
cuda_check(): API error 0715 (CUDA_ERROR_ILLEGAL_INSTRUCTION)
  in ext/drjit/ext/drjit-core/src/cuda_core.cpp:236
Fatal Python error: Aborted
  prbvolpath_sm.py, line 738 in _flush_deferred_probes
  prbvolpath_sm.py, line 718 in sample
```

Line 738 is `dr.compress(...)`, the first blocking readback after the replay
loop, so the faulting kernel is almost certainly the loop itself rather than
the compress.

Identical on Ubuntu 22.04 / Clang 17 and Ubuntu 24.04 / GCC 13, and identical
again after rebasing onto `e04bebb1` (Dr.Jit v1.5.0-6 to v1.5.0-20, nanobind
3.0.0-dev2), so it is deterministic on that agent. macOS and Windows are
green. The earlier Windows failure was `sccache: Address in use` and cleared
on its own.

**What I ruled out**, all on the rebased branch:

| check | result |
|---|---|
| `dr.set_flag(dr.JitFlag.Debug, True)` on the whole test file | no bounds violation |
| `compute-sanitizer --tool memcheck` on test06 | 0 errors |
| whole suite in one process, as CI runs it | 3345 passed, no abort |
| A40 (sm_86), A100 (sm_80), H200 (sm_90) | pass on all three |
| test06 x30 with a warm, persistent kernel cache | 0 failures |
| 1 to 9 live `Medium` instances during the adjoint pass | no change |

So it does not reproduce on any hardware I have.

**What I would need.** Which GPU is on the agent, and ideally one run of that
test with `DRJIT_LOG_LEVEL=4` or `dr.set_flag(dr.JitFlag.Debug, True)`. If
you would rather not spend agent time on it, I am happy to push a temporary
commit that turns those on so the CI prints them, and drop it afterwards.

Worth noting what is specific to this test: it is the only one that puts a
three-channel extinction grid in an RGB variant, and the record buffer it
fills is written with `dr.scatter_inc` inside a symbolic loop during the
adjoint pass. If that combination is known to be fragile on older
architectures, that would narrow it down quickly.

wjakob:
Hello Ruihan,

thank you for this PR. I am excited to ship this feature as part of Mitsuba. However, there are some aspects of the current realization in a PR that will need further work:

- the PR message and some parts of the commits look like they came straight out of Claude code. Generally direct output from LLMs is below the quality bar of this project.
- It's a big change at +2.5K LOC
- Part of why it is so big is that you introduce unrelated functionality for supergrids. I think that should really be factored out in its own PR so that we can discuss what a good interface could be. For example right now you baked DDA traversal into the bottom-level Medium class, but the bottom-level classes are designed to be very abstract to accommodate many possible implementations. For example, some users of Mitsuba (Rayference/Eradiate) work with curved grids for planetary data. If we bake a rectilinear version as the core interface, it will be completely incompatible with their workflow.

Best,
Wenzel

AuroraRyan0301:
Hi Wenzel,

Thanks for the feedback, and sorry for the late reply. I apologize for leaving so much AI-generated content in the PR. I just wanted to get the discussion about the PR and its design started quickly. While porting the implementation to Mitsuba 3.9, I mainly focused on supporting a broad range of variants, including spectral rendering, and achieving good performance.

I have now split out the features and simplified the commits and PR description. I have also removed the supergrid functionality for now, since making it compatible with Eradiate seems to require substantial changes to the interface. I hope the revised code is closer to the project's quality standards.

Thanks again for your time and support.

Best,
Ruihan

PR #1947, opened 2026-09-04
Try merging from stable branch instead of master

andrewgait:
This option didn't appear to be available last time I looked at this - let's see if it helps at all.


andrewgait:
Sorry - this isn't what I was trying to do - closing.

PR #1948, opened 2026-09-06
Improve bounding box Python bindings

dvicini:
This PR improves the bounding box bindings:

1. It adds bindings for BoundingBox1f, which are used to represent a time range for animation support. 
2. It fixes how `ray_intersect` is bound for scalar bounding boxes. Previously, the definition of this binding depended on import order. Depending on which variant was imported first, `ray_intesect` on the scalar bounding box accepted a different type for Ray (scalar or LLVM or CUDA ....). Now, we make sure to consistently add all overloads for this function.
3. The PR also adds some basic unit tests to exercise both 1D and 2D bounding box bindings

### DIFF
--- a/src/core/python/bbox_v.cpp
+++ b/src/core/python/bbox_v.cpp
@@ -86,12 +86,32 @@ template <typename BBox, typename Ray> auto bind_bbox(nb::module_ &m, const char
     }
 }
 
+// Add a `ray_intersect` overload to an existing bounding box binding.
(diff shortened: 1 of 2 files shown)


wjakob:
Thanks!

wjakob:
> Now, we make sure to consistently add all overloads for this function.

I don't think it matters (even if a overload shows up twice). Nanobind will only pick exact matches when traversing overloads on the first try.

PR #1949, opened 2026-09-06
Fix BoundingSphere3f Python bindings

dvicini:
This fixes two issues with the BoundingSphere3f Python bindings:

1. Calling `bounding_sphere()` on a bounding box returns a bounding sphere by value. Because bounding sphere was inheriting from `drjit::TraversableBase`, this did not work and crashed. This PR makes bounding sphere mirror the patterns of bounding box to make it a proper value type.

2. Similar to #1948 for bounding boxes, this also fixes the binding of the `ray_intersect` function to not depend on import order.


wjakob:
Thanks!

PR #1951, opened 2026-09-07
Improvement pass over BSDFs (esp. `principled`, `principledthin`)

wjakob:
This PR contains an AI-assisted pass over the `principled` and `principledthin` BSDFs to fix a number of lingering problems and inefficiencies. Where applicable, it also generalizes those fixes to other BSDFs in Mitsuba:

It addresses the following problems (roughly in order of decreasing severity):

- Lobe sampling used fixed weights that ignored the material parameters and incident direction. Sampling now adaptively chooses the lobes based on the expected amount of energy, which reduces noise in renderings using these models. The manual `*_sampling_rate` parameters no longer serve a purpose and were removed.
- The `BSDF::sample()` interfaces were implemented as `eval()/pdf()`, which performed several redundant texture evaluations. The implementation now only fetches the parameters once and computes the BSDF value and density together.
- The component indices reported by `sample()` were inconsistent with respect to the declared set of lobes.
 - The implementation relied on the compiler fusing multiplies and adds into FMAs, which is unspecified and backend-dependent. It now uses `dr::fmadd()` and `dr::lerp()` explicitly, reuses common subexpressions, and replaces square roots that only served as divisors with `dr::rsqrt()`. 
- The factor removes redundant masking, two duplicated copies of the GGX Smith shadowing function, and a doubled Schlick evaluation.
 - The implementation now precomputes more information when possible, such as IOR reciprocals and the reciprocals and normalization constants of the microfacet roughness. When the roughness parameters are spatially uniform, the entire microfacet distribution is built once and made opaque, so that kernels load these constants instead of recomputing them in every lane.


wjakob:
This commit improves geometric mean performance on a set of Blender test scenes by 1.22x. Together with the variance reduction, it's roughly an 1.4x improvement.

PR #1970, opened 2026-09-24
Fix GGX VNDF sampling at QMC boundaries

thecoober:
## Description

GGX visible-normal sampling can produce non-finite slopes when a QMC sampler generates exact boundary values such as `0` or `1`. The concentric square-to-disk mapping places these samples on singular disk-boundary configurations, where the subsequent slope conversion can divide by zero.

This change:

- maps the measure-zero boundary of the input sample onto the open unit square before the GGX VNDF mapping;
- applies a per-lane finite-value guard to the resulting slope;
- adds a regression test covering exact boundary samples on the scalar backend and each available vector backend.

No linked issue.

## Testing

The regression test was checked using a red-green procedure:

- with the unmodified GGX implementation, the new boundary test reports non-finite sampled normals or PDFs;
- after applying this fix and rebuilding Mitsuba, the boundary test passes.

The default Linux configuration successfully compiled the `scalar_*`, `cuda_ad_rgb`, `llvm_ad_rgb`, and `llvm_ad_spectral` variants.

Commands used:

```bash
ninja -C build -j8

python -m pytest src/render/tests/test_microfacet.py \
    -k boundary_samples_are_finite -vv

python -m pytest src/render/tests/test_microfacet.py -q

python -m pytest \
    src/bsdfs/tests/test_rough_conductor.py \
    src/bsdfs/tests/test_rough_dielectric.py -q
```

## Checklist

- [x] My code follows the style guidelines of this project
- [x] My code also compiles for `cuda_*` and `llvm_*` variants
- [x] No documentation changes are required for this internal numerical fix
- [x] I have added tests that prove my fix is effective
- [x] I cleaned the commit history and removed any merge commits
- [x] I give permission that the Mitsuba 3 project may redistribute my contributions under the terms of its license


thecoober:
These failures appear to be unrelated to this PR. I checked the TeamCity results for an upstream maintainer commit, and the same LUT and rendering tests fail there with the same errors.

